PR #2, opened 2026-08-19
feat: add agent-bundle compiler and developer workbench

ScriptedAlchemy:
Add the agent-bundle compiler, CLI, and developer workbench. RSC Runtime stays on draft #1. Native Claude/Codex smokes were not run.


ScriptedAlchemy:
Post-merge CI root cause: the relay-shell test added here kept `hostOrigin` hard-coded to `127.0.0.1:43123` while asking the proxy for `port: 0`. On Node 22.19 CI, the kernel legally assigned `43123`, so the production different-origin guard correctly rejected the fixture. Recent merges including #341 did not touch this path; #341's Rstest socket-root relocation is unrelated. #349 reserves an actual host listener, eliminating the collision while preserving the guard.

PR #37, opened 2026-08-29
test: fail the runtime capture fast with phase attribution instead of an opaque timeout

ScriptedAlchemy:
## Summary
- Root-causes the `Verify (Node 24)` failure on the #35 merge commit (run 33266629560): `runtime-playground-capture.test.ts` hit its 600s budget with **no output at all**, because the capture child process has bounded browser waits but no overall deadline, and its cleanup path (`restore`/`browser.close`/`fixture.close`) is unbounded — any wedge is an opaque outer-test timeout.
- Adds an in-process 480s watchdog (unref'd, main-module only) that exits 1 with the current capture phase on stderr, so the calling test's `execFile` rejects with attribution well inside its budget.
- Adds phase markers across the capture stages and bounds each cleanup step at 30s (recorded as a failed step instead of holding the process open), with the step budget injectable through the existing cleanup-boundary test seam.

## Test plan
- [x] `pnpm lint`, `pnpm typecheck`
- [x] Capture test file (all 3 tests, incl. full browser capture and the new wedged-cleanup-step test) green locally
- [ ] CI green

### DIFF
--- a/packages/workbench/scripts/capture-runtime-playground.mjs
+++ b/packages/workbench/scripts/capture-runtime-playground.mjs
@@ -8,7 +8,41 @@ import { timeScale } from '../../agent-bundle/tests/support/time-scale.ts';
 import { startRuntimePlaygroundFixture } from '../tests/helpers/runtime-playground-fixture.ts';
 
 const browserTimeout = 30_000 * timeScale;
+/**
+ * Hard ceiling for the whole capture, comfortably under the calling test's
+ * 600s budget so a wedge fails HERE with the current phase on stderr instead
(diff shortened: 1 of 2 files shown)


ScriptedAlchemy:
Coordination note from the sibling session: #36 just landed the same script hardening (watchdog with phase attribution, bounded cleanup steps, forced exit) for the main CI failure. The unit test here for the bounded cleanup is the piece #36 lacks — worth rebasing this PR down to the test (adapting to the merged implementation, which reads the step budget from a module constant rather than an injected parameter — an injectable parameter like this version has is arguably better for testability, so keeping your signature and adjusting the merged script would also work).

Issue #96, opened 2026-08-31
Build a real RSC Agent Document renderer with streaming semantics


ScriptedAlchemy:
## Summary

Replace the current synchronous React-element lowerer with an actual React
Server Components execution path. Agent routes render a transport-neutral Agent
Document rather than HTML. One live internal Flight stream carries async Server
Component and Suspense work to an Agent Document decoder; MCP, CLI, scripts,
Workbench, and tests then project the same document according to their real
capabilities.

This issue owns rendering and projection. Typed request context comes from the
request-context feature, cross-request state comes from the optional state
kernel, and messages for future or different invocations come from the directed
notices feature.

## Problem

`@agent-bundle/rsc-runtime` currently provides a small `Mcp.*`/`Hook.*` JSX
vocabulary and synchronous lowerers. The lowerers call function components as
ordinary functions, walk the resulting elements, and produce one completed
protocol object. CLI output bypasses JSX and serializes the validated operation
result directly.

That is a useful React-shaped protocol DSL, but it is not a React renderer or
an RSC/Flight runtime. React does not own component execution, so async Server
Components, Suspense, request cancellation, boundary errors, backpressure, and
incremental reveal are unavailable. This is the implementation gap described
by #88.

## Goals

1. Run route components through React's RSC/Flight implementation rather than
   manually invoking component functions.
2. Define one immutable Agent Document vocabulary for agent-facing results.
3. Preserve a live Flight stream through decoding and projection.
4. Distinguish Suspense boundary replacement from mutable progress updates.
5. Produce honest MCP, CLI, script, Workbench, and test projections.
6. Pin and test one exact compatible React/RSC package set because framework
   integration APIs are not semver-stable across React minors.

## Non-goals

- Owning request identity or provider composition; that belongs to the request
  context feature.
- Treating React state, context, or `useSyncExternalStore` as server-side
  persistence.
- Keeping a completed Server Component mounted as a subscription.
- Inventing rich partial MCP tool results where the protocol exposes only
  progress notifications and one final result.
- Shipping Server Functions or browser actions in the first renderer release.
  The compiler must preserve client/server boundary information, but action
  transport is a separate feature.
- Parsing every Markdown string into the core document tree.

## Rendering architecture

```text
generated route entry
  -> install typed Agent request context
  -> render Server Component model to a live Flight stream
  -> incrementally decode Flight
  -> commit Agent Document snapshots/events
  -> target projector
       |- MCP progress + final CallToolResult
[...]

Issue #93, opened 2026-08-31
Compile filesystem routes into the Agent Bundle meta-framework


ScriptedAlchemy:
## Summary

Agent Bundle already owns configuration normalization, host artifact generation,
Rslib builds, Rsbuild-based MCP App compilation, generated lifecycle shells,
virtual modules, Workbench development, and Rstest verification. The missing
framework feature is the application compiler between project source and those
existing build/runtime layers.

Today, framework mode discovers package-level entries such as `src/cli.ts` and
`src/mcp/<server>.ts`, but individual MCP tools, resources, prompts, CLI
commands, and rendered operations are assembled manually at runtime. Projects
must import every operation, concatenate registries, repeat server identifiers,
and write thin MCP/CLI entry modules.

Add a convention-driven route compiler. The framework supplies standard source
roots and infers servers and route kinds from paths. Root configuration is only
for project-wide policy and explicit overrides. A route is one module:
its filesystem path supplies stable identity, its optional named `config` export
adds protocol metadata, and its remaining exports provide schemas, execution,
and React rendering. The compiler validates the complete route graph, generates
types and virtual registries, and feeds the existing Rslib/Rsbuild build paths.

This deliberately supersedes the prior `2026-08-26` design decision that listed
a filesystem router as a non-goal.

## Existing capabilities to preserve

This issue does **not** propose rebuilding capabilities that already exist:

- `agent-bundle` and `create-agent-bundle` are built with Rslib.
- Workbench and browser-facing MCP Apps are built with Rsbuild.
- The repository already uses Rstest for unit, integration, browser, packed,
  and native-smoke coverage.
- Explicit config already overrides conventional package entries.
- Generated lifecycle shells already own stdio startup, stdout protection,
  cancellation, shutdown, and heartbeat behavior.
- Generated modules already ride the Rslib build through protected virtual
  module entries.
- MCP Apps already compile into self-contained HTML and a generated registry.
- Project snapshots and the development watcher already detect source additions,
  edits, and deletions.

The compiler should extend these mechanisms rather than create a parallel build
system or expose Rspack as a product API.

## Problem

The current Audiobook Curator illustrates the gap:

- `agent-bundle.config.ts` declares a server named `curator`.
- `src/application.ts` manually imports five operation families and joins their
  arrays.
- each operation repeats its MCP server assignment and listing metadata;
- `src/mcp/curator.ts` repeats the server name and calls
  `createRscMcpServer`;
[...]

PR #113, opened 2026-09-01
feat: framework request store and agent() (#95)

### LINKED ISSUE #95: Add typed Agent request context and pluggable context providers
## Summary

Give every generated event route, tool, CLI command, rendered script, Server
Function, and browser-backed render request one typed, request-scoped Agent context. Let
framework and application providers contribute capabilities to that context so
authors can build APIs such as `useWorktree()`, `useAgentTopology()`, or
`useTraceDecay()` without modifying compiler core.

This issue provides the composition boundary. It does not provide durable
state, infer agent relationships, or require a daemon.

## Problem

The current `createRscRequestContext()` is a thin `AsyncLocalStorage` wrapper.
The full RSC example hand-builds `useEdit()` and `useRuntimeSnapshot()` around
it. That proves request scoping, but every application must still invent:

- the canonical request identity;
- host, session, thread, actor, workspace, and worktree availability semantics;
- service and storage injection;
- provider ordering and failure behavior;
- generated TypeScript augmentation;
- test fixtures and request-lifetime validation.

React Context is not available to Server Components, and client state hooks do
not provide server request scope. A framework-owned async context is required.

## Public model

```ts
interface AgentRequestContext {
  invocation: AgentInvocation;
  host: Observed<AgentHostIdentity>;
  session: Observed<AgentSessionIdentity>;
  actor: Observed<AgentActorIdentity>;
  workspace: Observed<AgentWorkspaceIdentity>;
  capabilities: AgentRequestCapabilities;
  progress: AgentProgressReporter;
  signal: AbortSignal;
  services: AgentServiceRegistry;
  providers: AgentProviderValues;
}

type Observed<T> =
  | { state: 'available'; value: T; source: 'native' | 'receipt' | 'derived' }
  | { state: 'unavailable'; reason: AgentContextUnavailableReason };
```

Unknown identity is never represented by a fabricated string or an ambiguous
optional field.

The capability set includes explicit server-side filesystem, command, network,
and project-root authorities where the invocation grants them. An RSC route may
therefore read files or run commands directly through those capabilities; it
does not need an MCP tool call merely because it executes on the server.

Async Server Components and ordinary server utilities use:

```ts
const context = await agent();
```

The RSC renderer issue adds a synchronous `useAgent()` convenience by unwrapping
one stable request promise through React `use()`. The core request store remains
React-independent and available to ordinary server utilities.

## Context-provider extension

```ts
interface AgentContextProvider<Name extends string, Value> {
  readonly name: Name;
  readonly requirements?: AgentProviderRequirements;
  resolve(request: AgentRequest): Promise<Value>;
}
```

Providers may be enabled project-wide in `agent-bundle.config.ts` or discovered
through a conventional `src/providers/` root. The compiler:

- validates unique provider names and dependency order;
- generates the augmented context type;
- records provider requirements in the semantic manifest;
- rejects provider cycles;
- prevents request-scoped values from entering persistent caches;
- fails closed when a required host event or capability is unavailable;
- gives Rstest a provider fixture boundary.

Application code can define an async provider accessor without framework
changes:

```ts
export async function worktree(): Promise<WorktreeContext> {
  return (await agent()).providers.gitWorktree;
}
```

After the renderer's React bridge exists, a synchronous custom `useWorktree()`
may delegate to `useAgent()`. Async components should still prefer
`await worktree()` or `await agent()`.

Providers resolve request-scoped capabilities. They do not implicitly observe
events or mutate durable state. An application that needs those behaviors uses
an explicit semantic event route and state transaction, keeping provider
composition independent from the optional state kernel.

## Request lifecycle

The framework installs versioned, realm-singleton async stores at every real
entrypoint. The context is read-only during `render`, explicitly mutation
capable during `action`, and reduced to captured values during `after`.
Access outside a real invocation throws a typed framework error.

Artifact epoch, source revision, host contract revision, and protocol revision
travel with the invocation so stale Server Function or Flight references are
rejected.

## Non-goals

- A built-in thread-topology algorithm.
- A mandatory local server or daemon.
- React Context as server state.
- `useSyncExternalStore` in Server Components.
- Inferring parent/subagent relationships absent from a host contract.

## Delivery steps

1. Promote `createRscRequestContext` into a framework-owned request store.
2. Define the canonical request and `Observed<T>` identity contracts.
3. Add `agent()` and synchronous `useAgent()`.
4. Add provider registration, ordering, requirement validation, and typegen.
5. Bind event-route, MCP, CLI, script, and Server Function entrypoints.
6. Add test provider fixtures and cross-request isolation tests.

## Acceptance criteria

- The same typed context is available across every generated server entrypoint.
- Request context survives `await` and is absent after invocation completion.
- A project-defined provider adds a typed context property without compiler
  modification.
- Two simultaneous requests cannot observe each other's principal, event,
  capabilities, or provider values.
- Unavailable host identity is typed and provenance-bearing.
- Framework-owned persistent caches never retain request context, and access
  through a captured handle after request completion throws.
- The API works without a daemon or durable state provider.

## Design references

- [Current minimal `createRscRequestContext` implementation](https://github.com/ScriptedAlchemy/agent-bundle/blob/560124af0e23719a9e7cb006ef5ddb94b401f749/packages/rsc-runtime/src/request-context.ts)
- [Current RSC example request-context usage](https://github.com/ScriptedAlchemy/agent-bundle/blob/560124af0e23719a9e7cb006ef5ddb94b401f749/examples/rsc-agent-runtime/src/runtime/request-context.ts)
- [Node.js `AsyncLocalStorage`](https://nodejs.org/api/async_context.html#class-asynclocalstorage)
- [React Server Components async rendering contract](https://react.dev/reference/rsc/server-components)
- [React `cache` is request-scoped, not durable storage](https://react.dev/reference/react/cache)

## Stack position

- Depends on: [#93 filesystem route/compiler IR](https://github.com/ScriptedAlchemy/agent-bundle/issues/93).
- Unlocks: [#96 RSC renderer](https://github.com/ScriptedAlchemy/agent-bundle/issues/96), [#97 semantic events](https://github.com/ScriptedAlchemy/agent-bundle/issues/97), [#98 durable state](https://github.com/ScriptedAlchemy/agent-bundle/issues/98), [#104 proximity example](https://github.com/ScriptedAlchemy/agent-bundle/issues/104), and [#105 Workbench](https://github.com/ScriptedAlchemy/agent-bundle/issues/105).
- Related motivation: [#88](https://github.com/ScriptedAlchemy/agent-bundle/issues/88).

## Full meta-framework stack

- Foundations: [#93 routes](https://github.com/ScriptedAlchemy/agent-bundle/issues/93) - [#94 identity](https://github.com/ScriptedAlchemy/agent-bundle/issues/94) - [#95 context](https://github.com/ScriptedAlchemy/agent-bundle/issues/95) - [#96 renderer](https://github.com/ScriptedAlchemy/agent-bundle/issues/96)
- Runtime: [#97 events](https://github.com/ScriptedAlchemy/agent-bundle/issues/97) - [#98 state](https://github.com/ScriptedAlchemy/agent-bundle/issues/98) - [#99 notices](https://github.com/ScriptedAlchemy/agent-bundle/issues/99)
- Host: [#100 components](https://github.com/ScriptedAlchemy/agent-bundle/issues/100) - [#101 lifecycle](https://github.com/ScriptedAlchemy/agent-bundle/issues/101)
- Surfaces and proof: [#102 CLI/scripts](https://github.com/ScriptedAlchemy/agent-bundle/issues/102) - [#103 Rstest](https://github.com/ScriptedAlchemy/agent-bundle/issues/103) - [#104 proximity example](https://github.com/ScriptedAlchemy/agent-bundle/issues/104) - [#105 Workbench](https://github.com/ScriptedAlchemy/agent-bundle/issues/105)

ScriptedAlchemy:
## Summary

Wave 1 Lane C context core for #95.

- Versioned realm-singleton request store with closed leases.
- `await agent()` as the React-independent accessor.
- Canonical `AgentInvocation` carrying the invocation-kind discriminator (`tool` | `event` | `cli` | `script` | `workbench`).
- Observed identities with provenance; unavailable is a typed reason, never a fabricated string.
- Reserved `state` / `notices` / `providers` extension slots.
- MCP and CLI entrypoints install the store behind the current operations model.

## Proof

`examples/rsc-agent-runtime` deletes its hand-rolled `withRenderContext` / `useEdit` / `useRuntimeSnapshot` internals. The hook Flight worker installs `runAgentRequest` for every render, and `AfterFileEdit` reads `await agent()`. Nested capability lists (`roots`, `allow`) are copied and frozen so a caller cannot mutate authority during a live request.

## Remaining sequenced work (not this PR)

- Provider discovery and generated typing ride #93's route graph.
- Synchronous `useAgent()` rides #96's React bridge.

This PR does not implement durable state, the renderer `useAgent()` React bridge, or the filesystem route compiler.

## Test plan

- `packages/rsc-runtime` tests and typecheck
- Example tests stay green
- Topology check green after dropping the deleted helper
- Do not merge

Fixes #95


packages/rsc-runtime/src/agent-request.ts
```diff
@@ -0,0 +1,314 @@
...
+  return lease.handle;
+};
+
+export const runAgentRequest = async <T>(
+  init: AgentRequestInit,
+  operation: () => T | Promise<T>,
+): Promise<T> => {
+  const values: FrozenValues = Object.freeze({
+    actor: init.actor ?? unavailable<AgentActorIdentity>(),
+    capabilities: Object.freeze({ ...(init.capabilities ?? emptyCapabilities()) }),
```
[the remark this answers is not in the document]
ScriptedAlchemy: Fixed in 1d90fba9: available() now recursively copies and freezes plain objects and nested arrays, with a regression test covering retained roots/allow mutation.

packages/rsc-runtime/src/agent-request.ts
```diff
@@ -0,0 +1,323 @@
...
+    host: init.host ?? unavailable<AgentHostIdentity>(),
```
ScriptedAlchemy: **[P2] Snapshot every Observed input at the request boundary**

The new available() helper snapshots its value, but AgentRequestInit publicly accepts the structural Observed<T> type, so callers can pass a plain object and bypass that helper. I reproduced this with a plain available host: mutating the retained host value during the request changed the value returned by agent(), and the exposed context.host was not frozen. The same bypass applies to actor/session/workspace and to capability entries, including authority-bearing values. Normalize/freeze every supplied Observed value inside runAgentRequest (or make constructed observations opaque) so the request boundary, rather than caller discipline, enforces the read-only snapshot contract.

ScriptedAlchemy: Fixed in 8463d022: runAgentRequest now snapshots every Observed input (host/actor/session/workspace and each capability entry) so a plain object cannot bypass available().

ScriptedAlchemy:
Landing per owner authorization on scoped verification: @agent-bundle/rsc-runtime 34/34 plus typecheck, the RSC example pool 187 passed, and eval:spot green. Prior local gate requirements and hosted-check waiting are superseded; main CI and the end-of-Wave comprehensive pass are the safety net.

Issue #73, opened 2026-08-31
Replace raw Rsbuild HMR frames with a provider-owned Runtime App reload channel


ScriptedAlchemy:
Follow-up to the Rsbuild conformance audit: #72.

## Problem

`RuntimeClientSurfaceProxy` opens a raw WebSocket to Rsbuild and parses text JSON with a top-level `type`. Rsbuild publicly documents token acquisition, `dev.client.path`, and `Environment.hot.send('full-reload')`; it does not stabilize the serialized server-message envelope, `hash` / `ok` ordering, heartbeat behavior, or raw-client obligations. A minor upgrade can leave the socket connected and health counters green while Runtime App refresh silently stops.

The bounded conformance fix allowlists only the provider-emitted `full-reload` kind and ignores private/unknown kinds. That removes the current semantic error but does not make the raw envelope public.

## Proposed direction

- Emit a provider-owned reload event from the successful, changed App environment compile hook.
- Carry that event through a provider/runtime-session-owned channel to the Runtime App outer relay.
- Keep Rsbuild WebSockets only for Rsbuild's shipped HMR client, not as an application protocol.
- Preserve current reconnect, teardown, payload-bound, and stale-refresh protections.

## Acceptance

- Runtime App reload behavior has no dependency on Rsbuild internal frame names or payload shape.
- A changed App compilation refreshes once; RSC/widget-only and unchanged App compilations do not refresh.
- HMR E2E covers connection loss/recovery and upgrade-frame changes without inspecting private Rsbuild messages.

PR #114, opened 2026-09-01
fix(test): consume runtime-app-reload frames monotonically (#111)

### LINKED ISSUE #111: flake: overview.e2e sees an extra runtime-app-reload generation frame (1-in-5 on main, Node 26) — dominant local-ci gate tax
## Summary

`packages/workbench/tests/overview.e2e.test.ts :: offers the host-owned MCP playground handoff...` fails intermittently with an **extra `runtime-app-reload` generation frame** beyond what the test expects. Reproduced **1-in-5 on unmodified `main` under Node 26** inside a local-ci leg worktree on a loaded machine.

This is currently the dominant tax on the merge loop: at roughly 20% per leg across the three verify legs, a fully green `pnpm check:local-ci` run is close to a coin flip, and the last two landings (#86, #109) each burned 4-5 gate attempts on pre-existing load-sensitive flakes rather than regressions.

## Evidence

- Controlled repro during the #109 landing: 1 failure in 5 runs on unmodified main (`39590cb5` era), Node 26, same leg worktree, machine load in the 25-176 range. Failure mode is always one additional `runtime-app-reload` generation frame observed by the overview page assertion.
- Secondary flakes seen once each in the same gate runs (lower priority, listed for the pattern): `mcp-app-real.e2e.test.ts`, and `dev-watcher.test.ts` where chokidar coalesced a create event.
- Related: #110 tracks a separate cross-leg /tmp contamination failure in `cli.test.ts`; together these are the current sources of gate noise.

## Likely home of the defect

The generation frames come from the provider-owned Runtime App reload channel introduced by #73. Two candidate shapes:

1. The channel emits a duplicate/extra generation frame under load (e.g. a coalescing window that reopens, or an initial-state frame racing a real reload), in which case the channel should deduplicate by generation ordinal before broadcast.
2. The channel is behaving as designed (frames are at-least-once) and the test's exactly-once assumption is wrong, in which case the assertion should be relative/monotonic over generation ordinals rather than counting frames — the same fix shape PR #20 applied to `dev-provider.integration.test.ts`.

Deciding which requires capturing the frame sequence on a failing run (the SSE hub already carries sequence/replay-gap information that should make this cheap to log).

## Acceptance

- The failing assertion either consumes generation frames monotonically (order- and duplicate-tolerant) or the channel provably emits one frame per generation under induced load, with a regression test.
- 20 consecutive runs of the overview e2e on a loaded machine (load comparable to a 4-leg local-ci run) pass on Node 22.19/24/26.
- Extend the existing HMR/dev e2e rather than adding a parallel test file (repo convention).

ScriptedAlchemy:
Fixes #111

## Problem

`packages/workbench/tests/overview.e2e.test.ts` (`offers the host-owned MCP playground handoff...`) failed intermittently with an extra `runtime-app-reload` generation frame beyond the exact HMR message list it pinned. Reproduced 1-in-5 on unmodified `main` under Node 26, and it was the dominant local-ci gate tax.

Captured green sequences on this machine were `[generation 0 (connect replay), generation 1 (changed App compile)]`. A later generation-0 frame appears only after MCP playground handoff opens a second proxy socket. The flake is an extra frame of an already-seen or late ordinal landing between the post-edit snapshot and config-reconcile equality.

## Fix

Two complementary shapes, both required by the evidence:

1. **Channel:** the #73 App reload plugin no longer mints a frame from hashless/empty-hash App completions. Treating the stats object identity as a hash emitted a spurious reload on every unidentifiable completion. Regression in `dev-provider.integration.test.ts` (`emits one owned App reload...`).
2. **Assertion (PR #20 shape):** overview e2e consumes generations monotonically (non-decreasing observed sequence, duplicate-tolerant unique ordinals as a prefix). An extra at-least-once frame or a late compile from the source edit no longer fails config-reconcile. Failures dump generation ordinals, raw frames, and SSE hub `{latestSequence, subscriptionCount}`.

Product contract is unchanged: one owned reload per later successful *changed* App compilation. Hashless compiles were never a proven generation.

## Test plan

- [x] Plugin unit: hashless/empty-hash completions do not emit
- [x] 20/20 consecutive overview handoff e2e runs on Node 22.19.0, loaded machine (loadavg 7.4-11.0 on 8 cores, `AGENT_BUNDLE_TEST_TIME_SCALE=4`)
- Node 24/26 not installed on this box (only v22.19.0 at `/home/box/.local/node/current` and system v20.19.2)

### DIFF
--- a/packages/workbench/tests/overview.e2e.test.ts
+++ b/packages/workbench/tests/overview.e2e.test.ts
@@ -362,16 +362,28 @@ e2e('offers the host-owned MCP playground handoff only after a selected Runtime 
     ]);
     // The owned reload channel carries provider-authored frames only; a
     // changed App compile advances the generation past the connect replay.
+    // Frames are at-least-once: the proxy replays the current ordinal on
+    // connect, and one source change may burn more than one ordinal before
+    // the App compile coalesces (same shape as #20). Consume unique ordinals
+    // monotonically instead of counting frames.
     const ownedReloadFrames = (): readonly number[] => runtimePreviewHmrMessages.flatMap((message) => {
       try {
         const parsed = JSON.parse(message) as Readonly<{ readonly generation?: unknown; readonly kind?: unknown }>;
-        return parsed.kind === 'runtime-app-reload' && typeof parsed.generation === 'number' ? [parsed.generation] : [];
+        return parsed.kind === 'runtime-app-reload' && typeof parsed.generation === 'number' && Number.isSafeInteger(parsed.generation)
+          ? [parsed.generation]
+          : [];
       } catch {
         return [];
       }
     });
+    const ownedReloadOrdinals = (): readonly number[] => [...new Set(ownedReloadFrames())].sort((left, right) => left - right);
+    const expectMonotonicReloadOrdinals = (ordinals: readonly number[]): void => {
+      expect(ordinals.every((generation, index) => index === 0 || generation > ordinals[index - 1]!)).toBe(true);
+    };
     await expect.poll(() => ownedReloadFrames().some((generation) => generation > 0), { timeout: browserTimeout })
       .toBe(true);
+    expectMonotonicReloadOrdinals(ownedReloadOrdinals());
+    expect(Math.max(0, ...ownedReloadOrdinals())).toBeGreaterThan(0);
     const refreshedWidget = async () => {
       for (const frame of page.frames()) {
         if (await frame.getByTestId('runtime-hmr-marker').count() === 1) return frame;
@@ -391,7 +403,7 @@ e2e('offers the host-owned MCP playground handoff only after a selected Runtime 
     expect(runtimeAppRequests.filter((request) => request === 'POST /api/runtime/apps')).toHaveLength(1);
     expect(runtimeAppRequests.filter((request) => request.startsWith('DELETE /api/runtime/apps/'))).toHaveLength(0);
     expect(runtimeAppResponses).toHaveLength(1);
-    const hmrMessagesBeforeConfigReconcile = [...runtimePreviewHmrMessages];
+    const reloadOrdinalsBeforeConfigReconcile = ownedReloadOrdinals();
     const runtimeAttribute = async (name: string): Promise<string> => {
       const value = await runtimeIdentity.getAttribute(name);
       if (value === null) throw new Error(`Runtime identity omitted ${name}.`);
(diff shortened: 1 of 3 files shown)


packages/workbench/tests/overview.e2e.test.ts
```diff
@@ -415,7 +434,9 @@ e2e('offers the host-owned MCP playground handoff only after a selected Runtime
       await expect(runtimeIdentity).toHaveAttribute('data-runtime-source-revision', sourceRuntimeIdentity.sourceRevision);
       await expect(runtimeIdentity).toHaveAttribute('data-runtime-state-version', sourceRuntimeIdentity.stateVersion);
       expect(runtimePreviewHmrSockets).toHaveLength(1);
-      expect(runtimePreviewHmrMessages).toEqual(hmrMessagesBeforeConfigReconcile);
+      expectMonotonicReloadFrames();
+      const reloadOrdinals = ownedReloadOrdinals();
+      expect(reloadOrdinals.slice(0, reloadOrdinalsBeforeConfigReconcile.length)).toEqual(reloadOrdinalsBeforeConfigReconcile);
```
ScriptedAlchemy, line 439: **[P2] Reject reload ordinals minted during config reconciliation**

This prefix assertion allows arbitrary new ordinals after the saved prefix. If the two initial watched-source edits coalesce into generation 1, then a spurious reload caused by the later finite, invalid, or repaired config write can become generation 2 and this helper still passes. That weakens the prior contract that those reconciliations retain the existing preview without emitting another App reload. First wait for the edit-triggered sequence to settle, snapshot it, and require the sequence to remain exactly unchanged throughout the config-reconcile phase (while still tolerating duplicate frames for already-seen ordinals).

ScriptedAlchemy:
Reconciled with #120 (Wave 1 lane 0), which supersedes this PR for issue #111; closing per the owner's instruction (branch left intact).

**Adopted from this PR, with credit:** the `runtimeAppReloadPlugin` change and its regression coverage, ported into #120 (commit b60da260). Verified in the plugin harness on unmodified main that the defect is real: a hashless success falls back to stats object identity, so it always mints (`[1,2,3]` → `[1,2,3,4]`) and clobbers the retained hash, making the next unchanged hashed completion mint a spurious frame too (`→ [1,2,3,4,5]`). One correction to this PR's description: the fix does not stop hashless completions from minting (your own regression asserts they still reload, correctly — unidentifiable is not proven unchanged); what it fixes is the object-identity fallback and the retained-hash clobbering. #120 adds a regression for that clobbering case (unchanged hash after hashless stays deduped), verified to fail against the unfixed plugin.

**Not adopted:** the overview e2e assertion shape. #120's captured failing sequence (`generation 0, 1, +2` with the compiler log showing the two watched-source writes splitting into two App compilations) shows the extra frame is a real new generation, so #120 bounds the maximum generation by the edit budget in addition to monotonic consumption — a config reconcile that reloads the retained App still fails once the edit generations are spent, which the unbounded prefix check here would let pass. #120 also carries the 20/20-per-line proof across four Node lines (22.19.0 / 22.23.1 / 24.19.0 / 26.8.1, load 37–92), covering the Node 24/26 gap noted in this PR's test plan.

PR #118, opened 2026-09-01
perf(test): remove the CI serial-worker pin with flakes fixed at the source

ScriptedAlchemy:
## Summary

Removes the CI serial-worker pin from the integration pool: CI now derives workers from cores exactly like local (2 workers on 4-core hosted runners), with every flake that surfaced during burn-in root-caused and fixed at the source rather than absorbed by the serial shape. Subsumes and supersedes #90 (`perf/rstest-auto-workers`), whose per-`RSTEST_WORKER_ID` isolation is merged here intact.

## Why the pin existed, and why it can go

The pin landed because early 2-worker matrix runs flaked on a rotating test per leg. Burn-in on the original failing shape (taskset to 4 cores, `CI=1`, 2 workers) reproduced and root-caused each rotation member. None of them was "parallelism is unsafe" — each was a specific race or an unscaled budget:

| Flake | Test | Root cause | Fix |
|---|---|---|---|
| Extra App-preview close | `mcp-page-app-browser.test.ts` | Frame relay's fixed 1s force-close timer misfires against a healthy graceful teardown on a loaded host, superseding the queued graceful close | Load-tolerant 5s default for `closeTimeoutMs` (`mcp-app-frame.tsx`); server graceful-close receipt window raised to 35s so it dominates the relay's 30s cap (`mcp-app-routes.ts`) |
| Descendant-readiness race | `script-playground-service.test.ts` | Fixed 100ms sleep raced child Node startups; wrapper scripts published PID files non-atomically (`Number('') === 0` probed the runner's own process group) | Readiness sequenced through the injectable process-tree seam; PID files published via staged renames; polling budgets follow `AGENT_BUNDLE_TEST_TIME_SCALE` |
| timeline-App POST timeout | `mcp-app-real.e2e.test.ts` | `waitForRequest`/`waitForResponse` and several `expect.poll` calls used fixed budgets unscaled by the suite time scale; socket observations asserted synchronously against CDP event delivery | All capture budgets scaled; socket assertions converted to polls |
| Extra HMR reload frame | `overview.e2e.test.ts` | Two simultaneous watched-file writes could split into two rsbuild compiles; the dev middleware holds asset requests during a compile, so the refreshed DOM can show both edits while the second reload announcement is still in flight, poisoning the captured baseline | Each edit is barriered on its own announced reload generation before the next write |
| Extra reconcile cycle | `dev-workbench.test.ts` | Torn prepared project: `loadConfig` and the source snapshot read the config in two separate reads, so a write landing between them produced a stale model under a fresh revision hash | `#prepare` hashes the config before load and after snapshot and restarts when they disagree (`project-service.ts`); watched config writes use staged renames |
| CLI suite timeout | `tests/cli.test.ts` | Pack/build/spawn budgets were fixed literals, ignoring the suite time scale | Budgets scaled by `timeScale` |
| Cold-dist build race | `runtime-playground-fixture.ts` consumers | Two workers on a cold tree raced the same ensure-build of `examples/rsc-agent-runtime/dist` | `globalSetup` builds the payload once in the orchestrator before workers start |

Each fix names its race; none inflates a timeout to paper over one (budget changes are scale-convention fixes or documented dominance requirements between paired timers).

## Reconciliation with #90

#90's `RSTEST_WORKER_ID` isolation (per-worker cache/tmp/pack roots, `setupFiles`, `isolate: true`, unpinned unit/packed/watch pools) is merged here unchanged — it is correct hardening and prevents cross-worker root collisions. It does not, however, address any of the three originally cataloged flakes: those live in product timers, readiness sequencing, and unscaled budgets, not in shared-root collisions. Verdict per flake: all three were **still-needs-fix** (fixed in this branch), none **fixed-by-#90**.

This branch is the single coherent landing: #90's isolation + the root-cause fixes + the integration-pool worker derivation the burn-in evidence actually covers. #90 should be closed with credit.

## Final worker derivation

Integration pool, CI and local alike: `max(1, min(4, floor(availableParallelism() / 2)))` — 2 workers on 4-core hosted runners. Rstest's own auto-sizing would pick cores − 1 (3 on hosted runners), but each worker drives a Chrome + dev-server + rsbuild pair, so halving keeps pairs from starving each other; the cap bounds memory on large machines; `AGENT_BUNDLE_INTEGRATION_MAX_WORKERS` overrides for bisection. Unit/packed/watch pools follow #90's unpinned auto-sizing.

## Evidence

Local burn-in, all on the CI shape (taskset 4 cores, `CI=1`, 2 workers), on a heavily shared 96-core host (load 40–90, swap full — substantially harsher than a hosted runner):

- Baseline on pinned `main`: 14 runs; reproduced the rotating flakes (catalog above).
- Post-fix waves at successive compositions: 12-run wave at `107b242d` clean except one overview HMR flake (root-caused, fixed in `9128116e`); 12-run wave at `9128116e` had 3 cold-window failures — two were the CLI fixed-budget defect (fixed in `45333196`), one was a scaled 60s assertion budget exhausted while four cold runs plus the full local-ci gate saturated the host simultaneously (remaining-budget analysis in the branch log; not a product race).
- A/B stress verification for each named fix (e.g. torn-prepare: 0/12 failures fixed vs. reproducible torn preparations unfixed; overview HMR: 0/10 fixed vs. 3/10 unfixed) under deliberate CPU contention.
- Definitive wave at the final tip (`45333196`) and a solo `pnpm check:local-ci` gate: results appended before merge.

Hosted evidence: this PR's own Verify matrix, plus rerun legs to accumulate ≥9 hosted legs where queue availability allows; if the hosted queue stalls >2h the local evidence above plus a green `check:local-ci` is the documented landing basis, with hosted proof accumulating post-merge.

## CI wall-time

Serial integration legs on hosted runners ran ~7–8 minutes; the 2-worker shape runs the same suite in ~4 minutes locally on equivalent cores (247–257s warm). Measured hosted before/after numbers will be reported from this PR's Verify runs.

### DIFF
--- a/packages/agent-bundle/src/dev/project-service.ts
+++ b/packages/agent-bundle/src/dev/project-service.ts
@@ -629,7 +629,7 @@ export class ProjectService {
     }
   }
 
-  async #prepare(command: ProjectCommand): Promise<PreparedProject> {
+  async #prepare(command: ProjectCommand, tornRetries = 0): Promise<PreparedProject> {
     const requestedRoot = resolve(this.#options.root);
     const registry = this.#registry;
     const requestedConfigPath = resolve(requestedRoot, this.#options.configPath ?? 'agent-bundle.config.ts');
@@ -664,6 +664,14 @@ export class ProjectService {
       return failedPreparation('AB7002', 'Unable to prepare project paths.', requestedConfigPath, 'project.invalid-source');
     }
     const configPath = resolve(root, this.#options.configPath ?? 'agent-bundle.config.ts');
+    const configIdentity = async (): Promise<string | undefined> => {
+      try {
+        return createHash('sha256').update(await readFile(configPath)).digest('hex');
+      } catch {
+        return undefined;
+      }
+    };
+    const configIdentityBeforeLoad = await configIdentity();
     log(this.#options.logger, 'project.load', { command, root });
 
     let loaded;
@@ -709,6 +717,24 @@ export class ProjectService {
       snapshot = await snapshotProjectSource(root, loaded.configPath, outputRoots, payloadRoots);
     } catch {
       return failedPreparation('AB7003', 'Unable to snapshot project source.', loaded.configPath, 'project.invalid-source');
+    }
+    // loadConfig evaluated the config from one read while the snapshot hashed
+    // it in another; a config replacement landing between the two reads would
+    // otherwise produce a torn preparation whose model belongs to the old
+    // bytes while its revision hashes the new tree. Consumers dedupe prepared
+    // deliveries by revision, so a torn preparation reconciles a stale model
+    // under a fresh revision. When the config changed mid-prepare, restart the
+    // preparation so both reads agree; the retry cap only yields once writes
+    // outpace prepares for several consecutive rounds, which no real editor
+    // or test harness sustains.
+    if (tornRetries < 3) {
+      const configIdentityAfterSnapshot = await configIdentity();
+      if (
+        configIdentityBeforeLoad !== undefined && configIdentityAfterSnapshot !== undefined &&
+        configIdentityBeforeLoad !== configIdentityAfterSnapshot
+      ) {
+        return this.#prepare(command, tornRetries + 1);
+      }
     }
     const runtime = runtimeDeclaration(this.#options.includeDevRuntime === true, loaded.config, loaded.configPath);
     const runtimeMetadata = runtime.declaration === undefined
--- a/packages/agent-bundle/src/dev/mcp-apps/mcp-app-routes.ts
+++ b/packages/agent-bundle/src/dev/mcp-apps/mcp-app-routes.ts
@@ -15,7 +15,11 @@ import type { McpAppConsentRequest } from './mcp-app-sandbox.ts';
 import { runtimeAppMessageLimits } from '../runtime-app-message-limits.ts';
 
 const bodyLimit = 64 * 1024;
-const gracefulCloseReceiptTimeoutMs = 5_000;
+// A force-close DELETE that lands after an accepted graceful close must stay
+// idempotent (200, not 404), so this window has to dominate the frame relay's
+// force-close budget — clients may fall back as late as their closeTimeoutMs,
+// which mcp-app-frame.tsx caps at 30s.
+const gracefulCloseReceiptTimeoutMs = 35_000;
 
 interface RequestDiagnostic {
   readonly code: string;
--- a/packages/agent-bundle/src/dev/mcp-apps/mcp-app-routes.ts
+++ b/packages/agent-bundle/src/dev/mcp-apps/mcp-app-routes.ts
@@ -88,6 +88,12 @@ export interface McpAppRoutePreviewService {
 
 export interface McpAppRoutesOptions {
   readonly authorize: (request: IncomingMessage) => void;
+  /**
+   * Test-only override for the graceful-close receipt window. Production
+   * callers must leave this unset so the window keeps dominating the frame
+   * relay's force-close budget.
+   */
+  readonly gracefulCloseReceiptTimeoutMs?: number;
   readonly service?: McpAppRoutePreviewService;
 }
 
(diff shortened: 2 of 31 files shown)


packages/agent-bundle/src/dev/project-service.ts
```diff
@@ -710,6 +718,24 @@ export class ProjectService {
...
+    // loadConfig evaluated the config from one read while the snapshot hashed
+    // it in another; a config replacement landing between the two reads would
+    // otherwise produce a torn preparation whose model belongs to the old
+    // bytes while its revision hashes the new tree. Consumers dedupe prepared
+    // deliveries by revision, so a torn preparation reconciles a stale model
+    // under a fresh revision. When the config changed mid-prepare, restart the
+    // preparation so both reads agree; the retry cap only yields once writes
+    // outpace prepares for several consecutive rounds, which no real editor
+    // or test harness sustains.
+    if (tornRetries < 3) {
```
ScriptedAlchemy, line 744: **[P2] Fail closed instead of publishing the fourth torn preparation**

After three consecutive config changes, tornRetries is 3 and this entire comparison is skipped, so the method proceeds with the config loaded before the latest write but the snapshot/revision captured after it—the exact stale-model/fresh-revision state this fix is intended to prevent. A save burst or generated config writer can sustain four rounds; assuming it will not does not preserve the preparation invariant. Keep checking on the final attempt and return a typed invalid/transient preparation (or otherwise retry outside this bounded call) when the identities still differ, rather than accepting the torn state.

ScriptedAlchemy:
## Verification against the #90 stop-findings (evidence transfer)

I resolved #90's conflicts and stopped it on two empirical findings ([evidence](https://github.com/ScriptedAlchemy/agent-bundle/pull/90#issuecomment-5487230200)). Verifying #118 against them, plus mergeability:

**a. Cold-dist cross-worker race — COVERED by this branch.** `rstest.integration.config.ts` adds `globalSetup: ['./rstest.integration.setup.ts']`, which runs `ensureRuntimeExamplePayload()` (extracted to `packages/workbench/tests/helpers/runtime-example-payload.ts`) once in the orchestrator before any worker starts, so no worker ever sees a cold `examples/rsc-agent-runtime/dist`. Re-verified empirically at the new head: `pnpm build`, `rm -rf examples/rsc-agent-runtime/dist`, then `test:integration:run` — 556 passed, 0 failed (the same fresh-tree procedure that failed twice on #90's branch and passed on main).

**b. Browser-setup defect — WAS MISSING, now pushed.** This branch had the same defect #90 had: `rstest.runtime-playground.browser.config.ts` and `rstest.runtime-playground.config.ts` loaded `rstest.setup.ts` (imports `node:fs`/`node:os`/`node:path` via `rstest.worker-isolation.ts`) into browser pools, where `node:` is an unhandled scheme — the standalone browser pool failed to build with 0 tests run. Cherry-picked the fix from #90's branch as 0298dae6: browser projects load an empty `rstest.setup.browser.ts`; the Node setup is scoped to the `runtime-node` project. Post-fix browser pool: 2 passed, exit 0.

**c. Mergeability with main — the branch was CONFLICTING (merge-base b0f14c8f predates #109/#112); merged in c14f05fe.** Same three-file conflict pattern as #90 plus `cli.test.ts`: every install site in `dev-workbench-packaging`, `public-api-packed`, and `rsc-runtime-optional-packaging` now spreads #109's `cachedNpmInstallArguments` (all six are consumer-contract installs) while keeping this branch's `installedEnvironment()` / `isolatedCommandEnvironment()` per-worker isolation; `cli.test.ts` keeps both sides' imports (`timeScale` + `cachedNpmInstallArguments`, both used in auto-merged body changes); `shared-pack.ts` auto-merged with both argument lists on top of the isolation plumbing.

Local verification at 0298dae6 (fresh worktree): typecheck, lint, `test:packed` (20 passed), `test:unit` (1781 passed), fresh-dist `test:integration:run` (556 passed), browser pool (2 passed) — all green. Gating after the rest of the wave lands, per landing order.

ScriptedAlchemy:
Merged by owner instruction ahead of the stuck hosted queue (Verify Node 24 sat queued >1h; all completed hosted checks were green: Verify 22.19.0/26, Release gates, Examples, micro-eval, dependency review). Hosted CI on main accumulates the post-merge evidence per this PR's documented plan. Merge basis: squash of 88313ac5 (clean merge of main @ e8908cc2, verified clean against 847989d6 via merge-tree) with local verification at 0298dae6: typecheck, lint, test:packed 20 passed, test:unit 1781 passed, fresh-dist test:integration:run 556 passed, browser pool 2 passed.

PR #153, opened 2026-09-01
feat(runtime): Effect v4 foundation — vendored source, boundary module, conventions (#wave 3.5 stage 0)

ScriptedAlchemy:
## Summary

Wave 3.5 stage 0: the Effect v4 foundation plus the agentic feedback loop, with zero behavior change to any published entry.

- **Dependency**: exact-pinned `effect@4.0.0-rc.112` on `@agent-bundle/runtime` only. The brief named `rc.113`, which is not on the registry (latest published RC is `rc.112`); the deviation is documented in `docs/effect-conventions.md` for the re-pin chore. Nothing imports it from an entry graph yet — verified no `effect` code lands in `dist/{index,state,state/sqlite,plugin}.js`.
- **Vendored source**: `repos/effect` via git subtree (Effect-TS/effect `main`, squashed). Editor excludes in `.vscode/settings.json` (search/files/watcher/auto-import); `repos/**` also excluded from rslint (`globalIgnores`) and the root tsconfig. Test-pool and pnpm-workspace globs are root-anchored (`packages/*`, `packages/**/tests/**`) so they never swept `repos/` in.
- **AGENTS.md**: "Vendored repos" section — read-only, read `repos/effect/LLMS.md` (verified it exists) + agent-patterns before writing Effect code, never import from `repos/**`.
- **Agent patterns**: `agent-patterns/effect-{stream,scope,concurrency,errors}.md`, written from the vendored v4 source/tests, each < 70 lines.
- **Boundary module**: `packages/rsc-runtime/src/effect/boundary.ts` is the only legal `Effect.run*` site: `runPromise(effect, { signal })`, `interruptWhenAborted` (AbortSignal→interruption), `scopedAbortSignal` (interruption→AbortSignal), and cause mapping onto the existing typed errors (`AgentRequestError`/`AgentContractError` by instanceof, `AgentStateError` by name so the state kernel stays off the root graph; interrupt-only causes → `AbortError`). Unit-tested: abort mid-effect interrupts, interruption rejects as `AbortError`, typed failures round-trip identically.
- **Boundary enforcement**: custom rslint rule `effect-boundary/no-ad-hoc-run` (rslint supports TS plugin modules — chosen over the grep stand-in), banning `run*` member calls and named imports outside `src/effect/boundary.ts`, plus unit tests of the rule.
- **Language service**: `@effect/language-service` wired in `tsconfig.base.json` with `outdatedApi`/`serviceNotAsClass` as warnings. Wired manually rather than `npx @effect/tsgo setup` because setup's `prepare` patch mutates the workspace `typescript` on every install; noted in the conventions doc.
- **Cold-start baseline** (stage-2 budget gate, generated claude SessionStart stdio hook, 7 bare-process runs, Node v22.23.1): median **39.74 ms**, min 36.41, max 43.06. Recorded in `docs/effect-conventions.md` + `docs/effect-cold-start-baseline.json`; re-check via `pnpm bench:hook-cold-start -- --check`.
- **Changeset**: patch on `@agent-bundle/runtime` (public dependency metadata changed).

## Verification (scoped per landing policy)

- `rslint .` green; negative test confirmed the boundary rule reports violations
- root + runtime `typecheck` green
- full unit pool green (128 runtime tests incl. new boundary/lint suites)
- built artifacts byte-identical in behavior: no entry imports `effect`


ScriptedAlchemy:
Codex P2s addressed in https://github.com/ScriptedAlchemy/agent-bundle/pull/164 (merged as `7a9887bed714030572dac9736f18d39f612959e6`).

**Recheck aborted state when the Effect starts.** Both boundary bridges now share the same pattern: `abortToInterrupt` defers the check with `Effect.suspend`, registers the `abort` listener first, then rechecks `signal.aborted` immediately after `addEventListener` so a signal that aborts between composition and run interrupts instead of hanging. Regression coverage is in both `effect-boundary.test.ts` files (abort-after-composition-before-start).

**Track aliased Effect namespaces.** `effect-boundary/no-ad-hoc-run` now records `import { Effect as Fx }` / `import * as E from 'effect'` (and `Runtime`) and flags `Fx.runPromise` / `E.runSync` / siblings. Negative fixture: aliased `E.succeed` is still allowed.

PR #473, opened 2026-09-03
chore(examples): use built-in framework surface over hand-rolled setup

ScriptedAlchemy:
Fixes #51, #52 and #54 (daemon side; the client-side `MAKEFLAGS` transport item in #54 is handled separately).

## #51 — kill honoured while parked; batch, ETA and settlement hardening
- `processJob` now races the job's kill signal against the load gate + `admission.withPermits(1)`. A kill on a job parked at the gate or on the permit settles it `killed` (`killed while queued`) immediately via the existing `finishKilledBeforeRun` path, releases any heavy claim, and no longer blocks the rest of its lane. A kill that lands after `claimStart` is left to the executor (the race arm parks instead of interrupting an admitted run).
- `foldBatch` skips candidates whose state is not `queued`; `processLaneJob` re-checks state before the batch window sleep and before folding, so a kill-requested head neither waits for nor leads a batch.
- Queue ETA uses `remainingEstimateMs` (clamped at 0) and counts a head parked at the gate (new `Lane.head`, set from take to settlement). The existing assertion in `daemon-broker.test.ts` that encoded the negative-cancel bug (`waitEtaMs === 0`) now expects the queued job's estimate.
- `settleJob` isolates every step after `claimSettlement` (`markFinished`, metrics, `notifyWaiters`, `completeExit`, `settleAttachments`) with `settlementStep` (log + continue on defect). `settleAttachments` and `finishAttachment` apply the same per-step isolation.

## #52 — attachment registration races
- `replayThenGoLive` takes the replay snapshot and clears `pendingLive` in one `Effect.sync`, so a chunk emitted between registration and replay is delivered once.
- `completeAttachRegistration` re-checks `leader.attachments.get(ticket) === attachment` in a sync frame before any ledger write and returns early if the leader settled / the attachment was killed or released in the window. `updateRunning` / `updateAttached` carry `AND status NOT IN (<terminal>)` and record a transition only when a row changed.
- Early follower release from the stdout pump (`releaseSatisfiedAttachments`) is guarded: a ledger/metric defect is logged instead of surfacing as `pump-failed` and terminating cargo.
- `mergeStderr` is part of identity/coverage attach compatibility for raw-output leaders (`channelsCompatible`); demux runs never merge so the flag is irrelevant there. Batch folding is documented as unchanged (folded followers receive the composite's channels as produced).

## #54 — jobserver vs `CARGO_BUILD_JOBS`
- New `isSharedJobserverArmed()`; `cargoExecEnv` injects `CARGO_BUILD_JOBS` only when the FIFO is **not** armed, so an armed daemon lets `sharedJobserverDelta` inject `MAKEFLAGS` and cargo joins the shared budget. Caller `-j` / `CARGO_BUILD_JOBS` still win.
- README: parallelism paragraph, capability row and `CARGO_HAULER_JOBS_GRANT` env row state the precedence; `DaemonConfigShape.jobsGrant` doc updated.

## Tests (each written failing first, then fixed)
- `tests/daemon-races.test.ts` (new, in-process broker over the fake cargo via the shared `tests/broker-fixture.ts`, extracted from `daemon-async.test.ts`):
  - kill of a job parked on the permit settles `killed` within the await budget while the holder keeps running
  - kill-requested pending job is not folded into a batch (leader chosen via a rider so the doomed job is the fold candidate)
  - settlement completes (exit callback fires, directory entry cleared, successor runs alone) when `markFinished` dies
  - late attacher receives each of 40 replayed ticks exactly once with `markAttached` delayed 400 ms
  - follower whose `markAttached` is held until after the leader settled stays `done`
  - early follower release with a dying `markFinished` leaves the leader `done` (not `pump failed`) and still delivers the follower's exit
  - identity attach on `cargo test` requires equal `mergeStderr`
- `tests/daemon-ledger.test.ts`: late `markAttached`/`markRunning` never reopen a `done` row or add transitions.
- `tests/daemon-jobs-grant.test.ts` (new): `cargoExecEnv` matrix + a daemon spawn asserting `CARGO_BUILD_JOBS` unset and `MAKEFLAGS` carrying `--jobserver-auth=fifo:<stateDir>/jobserver.fifo`.

## Gate
`pnpm check` green locally: validate, build, typecheck, Effect diagnostics (pre-existing suggestions only), rstest 637 passed / 0 failed, route-unit 32 passed / 0 failed.

## Review status
_(filled in per AGENTS.md once the automated review runs)_

### DIFF
--- a/examples/audiobook-curator/package.json
+++ b/examples/audiobook-curator/package.json
@@ -32,7 +32,6 @@
   },
   "dependencies": {
     "@agent-bundle/runtime": "workspace:*",
-    "@modelcontextprotocol/server": "2.0.0",
     "react": "19.2.8",
     "zod": "4.5.4"
   },
--- a/examples/rsc-agent-runtime/src/dev/generation-materializer.ts
+++ b/examples/rsc-agent-runtime/src/dev/generation-materializer.ts
@@ -18,7 +18,7 @@ import type {
 } from '../runtime/contracts.js';
 import type { DevRuntimePreparedProject } from '../../../../packages/agent-bundle/src/dev/runtime-provider.ts';
 import type { DevRuntimeMcpServerDescriptor } from '../../../../packages/agent-bundle/src/dev/runtime-protocol.ts';
-import type { JsonObject, JsonValue } from '../../../../packages/agent-bundle/src/dev/types.ts';
+import type { JsonObject, JsonValue } from 'agent-bundle';
 import type {
   RuntimeGenerationActivationGuard,
   RuntimeGenerationAsset,
--- a/examples/rsc-agent-runtime/src/dev/provider.ts
+++ b/examples/rsc-agent-runtime/src/dev/provider.ts
@@ -1,6 +1,9 @@
-import type { DevRuntimeProvider, DevRuntimeStartContext } from '../../../../packages/agent-bundle/src/dev/runtime-provider.ts';
+import type { DevRuntimeProvider } from 'agent-bundle/api';
 
 import { RsbuildRuntimeSession } from './rsbuild-runtime-session.js';
+
+/** The start context the public provider contract hands `start`; the framework exports only the provider type itself. */
+type DevRuntimeStartContext = Parameters<DevRuntimeProvider['start']>[0];
 
 export const createDevRuntimeProvider = (): DevRuntimeProvider => Object.freeze({
   descriptor: Object.freeze({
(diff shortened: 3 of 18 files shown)


ScriptedAlchemy:
The two framework gaps listed under "Proposed framework issues" are now filed:

- #484 — `agent-bundle/test`: mount one state (and notice ledger) across several `renderRoute` calls; cites the three re-implementations in worktree-proximity, host-test, and audiobook-curator.
- #485 — export the dev-runtime provider protocol types (`DevRuntimeStartContext`, `DevRuntimeSession`, `DevRuntimeInspectionEnvelope`, `RuntimeGenerationStore`, `RuntimeMcpRegistry`, …) from `agent-bundle/api`; lists the remaining `packages/agent-bundle/src/dev/**` import sites in rsc-agent-runtime.

Issue #592, opened 2026-09-05
Meta-framework architecture: one application graph, capability-driven projections, one execution kernel


ScriptedAlchemy:
## Decision

Agent Bundle should converge on one coherent meta-framework architecture:

> **one application, one route graph, one execution kernel, one compiler pipeline, one manifest, one composite artifact, many host projections.**

The product should feel like Next.js for agent-host plugins: authors express application intent through conventions and a small public API; compiler/runtime machinery owns host differences, packaging, transport, and projection.

This issue is the architectural umbrella for simplification work that cuts across #555 (composite root), #564 (browser/web), the route framework, execution/runtime plumbing, compiler profiles, Workbench, testing, and distribution.

This is not a request to rewrite working subsystems immediately. It establishes the direction so new work stops adding parallel mini-frameworks.

## Product boundary

Agent Bundle's job is:

> **Compile one typed agent application into native experiences across agent hosts.**

Core scope:

- filesystem-routed application structure;
- typed route contracts;
- shared execution context;
- host capability resolution;
- host projections;
- self-contained composite artifact;
- CLI / MCP / browser presentation of the same application;
- devtools, validation, testing, and distribution around that compiled application.

Do not turn core into a general agent orchestrator, workflow engine, ORM/database framework, deployment platform, LLM abstraction layer, or generic web framework.

---

## 1. Make the application graph the canonical source of truth

Today MCP tools/resources/prompts/apps, CLI routes, scripts, events/hooks, providers, layouts, skills, commands, and browser-facing behavior have partially separate discovery/planning paths.

Internally they should lower into one **Application IR**.

Illustrative shape:

```ts
interface ApplicationIR {
  identity: ApplicationIdentity;
  routes: RouteNode[];
  providers: ProviderNode[];
  middleware: MiddlewareNode[];
  assets: AssetNode[];
  capabilities: CapabilityRequirement[];
}

type RouteKind =
  | 'tool'
  | 'resource'
  | 'prompt'
  | 'app'
  | 'command'
  | 'script'
  | 'event'
  | 'page';

interface RouteNode {
  id: string;
  kind: RouteKind;
  source: SourceModule;
  input?: SchemaRef;
  output?: SchemaRef;
  runtime: 'node' | 'browser' | 'worker';
  capabilities: CapabilityRequirement[];
  metadata: Record<string, unknown>;
}
```

The exact TypeScript shape is not normative. The architectural requirement is.

### Required result

[...]

PR #588, opened 2026-09-05
feat(build): hold the package build's dist bundles to AB6005

ScriptedAlchemy:
## Rule

The npm package build's `dist` JavaScript is held to the same self-containment rule as the host packs. Every emitted `dist/**/*.js` and `*.mjs` bundle — `dist/bin/<name>.js`, its rendered-route Flight worker, the generated install bin, and the `lib` entry `dist/<name>.js` — is walked as an ES module before `dist` is published, and a bare import specifier that is not a Node built-in fails the build. There is no escape hatch: an import a `tools.rspack`/`tools.rsbuild` hatch keeps external (`output.externals`) used to survive in `dist` and be judged only by the prepack gate; it now fails the package build exactly as it already failed a host pack.

What stays valid: `node:` and bare built-in specifiers (`isBuiltin`), relative imports of emitted files, `import.meta`, `createRequire(import.meta.url)` of packed files. A literal dynamic `import()` counts like a static import (the walker already did this — confirmed, and the fixture covers it); a non-literal dynamic import is a finding. `.d.ts` output is not walked; declarations may still reference declared dependencies (`check-declaration-imports` covers the framework's own packages). The walk reads import specifiers: a `require`, `createRequire(…)(…)`, or `import.meta.resolve(…)` call is not an import and is outside `AB6005` in `dist` as in a host pack — see Self-review and #591.

## Code chosen: `AB6005`, reused

`AB6005` is documented as "Generated JavaScript import from `<path>` uses unsupported specifier `<specifier>`", with recovery "Bundle every JavaScript dependency into the artifact, then rebuild it." Its `generatedPath` is the only context it needs: the package build reports the consumer-visible path (`dist/bin/<name>.js`, `dist/index.js`), so the message reads correctly for `dist` without a sibling code. No new code allocated.

The existing walker is reused, not forked:

- `validate-artifact-modules.ts`: `validateJavaScriptModules` gains a `reportedRoot` option so a tree validated in its staging directory is reported under the path a consumer sees (`dist/…` rather than the stage-relative `bin/…`). `resolveJavaScriptImport` takes the reported importer separately from the artifact-relative one it resolves against.
- `package-build.ts`: after the invariant assertions and before `publishArtifact`, walks the staged tree with `validateJavaScriptModules` and throws `DiagnosticError` on any finding, so nothing is published and the stage directory is removed (the fixture asserts both).
[...]

PR #602, opened 2026-09-05
feat(build): AB6005 refuses bare require/createRequire/import.meta.resolve loads in emitted modules

### LINKED ISSUE #591: AB6005: also refuse bare createRequire/require/import.meta.resolve loads in compiled modules
Follow-up from the #588 self-review (reviewer finding 1, dismissed for that PR as out of scope).

`AB6005` (`packages/agent-bundle/src/build/validate-artifact-modules.ts`, `module-imports.ts`) walks the ES-module import graph: static imports, literal dynamic `import()`, and `import.meta`. It does not read `require("pkg")`, `createRequire(import.meta.url)("pkg")` (or a bound loader), `require.resolve`, or `import.meta.resolve("pkg")` with a bare non-built-in specifier. Rspack leaves those calls in the emitted bundle, so a compiled module — in a host pack or, since #588, in the package build's `dist` — can still load a package from the consumer's `node_modules` through one of them without an `AB6005` finding. The prepack gate (`pack-dependencies.ts`) already reads exactly those forms as `AB7014` evidence, so the two gates agree today (such a package must be declared under `dependencies`), but the artifact is not fully self-contained in the sense the "Generated plugin output" rule intends.

Confirmed on #588's tree: `es-module-lexer` reports only `node:module` and `import.meta` for `const load = createRequire(import.meta.url); load("left-pad")` and for `import.meta.resolve("right-pad")`.

Proposed: teach the `AB6005` walk the same literal `require` / `createRequire` / `import.meta.resolve` semantics `pack-dependencies.ts` already implements (move the scanner to a leaf module both import — `config/conventional-entry.ts` pattern — do not copy it), applied to compiled modules only (prebuilt payloads stay opaque), with a relative target resolved inside the tree and a bare non-built-in reported as `AB6005`. This changes host-pack validation too, so it needs its own decision and changeset (pre-1.0 `minor`), fixtures for literal and computed `require`, `createRequire` (direct and bound, aliased), and `import.meta.resolve`, and a check that no generated code (`cli-bins.ts`, `entry-shell.ts`, install bins) uses these forms for a bare package today.

Docs to update with it: `AGENTS.md` "Generated plugin output", `docs/diagnostics.md` (`AB60xx`, `AB7014` evidence paragraph), `website/docs/{en,zh}/guide/distribution/validation.mdx`, the `AB7014` recovery string in `pack-inventory.ts`.

ScriptedAlchemy:
Closes #591.

## Rule

Every emitted `.js`/`.mjs` module `AB6005` walks — a host-pack module, a package build `dist` bundle (`dist/bin/<name>.js`, its Flight worker, the install bin, the `lib` entry), and the framework-generated modules that get the full parse — is now read for the CommonJS-style loads Rspack leaves in emitted output as well as for its ES import records: `require("x")` and `require.resolve("x")`; `createRequire(…)("x")` and `createRequire(…).resolve("x")` with the factory written out, namespace-qualified (`Module.createRequire(…)`, `require("node:module").createRequire(…)`), or aliased (`import { createRequire as mk }`, `const { createRequire: mk } = …`); a loader bound to a name and called later (`const load = createRequire(import.meta.url); load("x")`, `load.resolve("x")`) — which is exactly the shim Rspack emits (`__rspack_createRequire_require`) and therefore what a `tools.rspack` `externalsType: 'node-commonjs'` external compiles to, closing the last externalization route #588 left open; and `import.meta.resolve("x")`. Prebuilt payload modules stay skipped (opaque consumer output), `.d.ts` is never walked.

Per load: a literal specifier `isBuiltin` accepts (`fs`, `node:fs`) passes; a literal relative or `file:` specifier is resolved with the existing `resolveJavaScriptImport` — exact listed regular `.js`/`.mjs` file, or listed valid JSON (host packs only; the package build passes `validJson: new Set()`), no CJS extension or directory probing — and the target module is then walked like an import target; a literal bare non-built-in is `AB6005` `uses unsupported specifier "left-pad" in <call>.`; a computed argument (`require(name)`, `require("driver/" + v)`, a template literal) is `AB6005` `loads a non-literal specifier through <call with …>.`; a loader passed on as a value rather than called (`const l = require`, `fn(load)`, `[require]`, `{ require }`, `x ? require : y`, `return load`, `=> load`) is `AB6005` `passes require on as a value instead of calling it.` / `passes load, a createRequire(…) loader, on as a value instead of calling it.`. `<call>` is the call as written — `require("left-pad")`, `require.resolve("left-pad")`, `createRequire(…)("left-pad")` / `mk(…)("left-pad")`, `createRequire(…).resolve("left-pad")`, `load("left-pad"), a createRequire(…) loader`, `import.meta.resolve("left-pad")` — and a relative-target failure is the resolver's existing text with the call appended (`is missing "./driver.cjs" in require("./driver.cjs").`, `resolves outside the artifact root: "../x.js" in import.meta.resolve("../x.js").`). Every message goes through the existing `graphDiagnostic`, so the prefix (`Generated JavaScript import from "<path>" `), code, severity, recovery (`Bundle every JavaScript dependency into the artifact, then rebuild it.`), and `generatedPath` (mapped under `dist/` by `reportedRoot`) are those of `AB6005` today; the import messages are unchanged byte for byte.

What never matches: a mention inside a comment, string literal, template literal, or regular-expression literal (the scanner steps over them as tokens — bundled library output is full of prose that says `require`); member calls (`host.require("x")`, `module.require`); private names (`this.#require(x)`); longer identifiers (`__webpack_require__("x")`, `require_fast_uri()`); `typeof require`; `path.resolve("x")`, `Promise.resolve("x")`; and a method or function definition named `require` (`require(id) {`).

#592 alignment: this extends the common scanner — `validateJavaScriptModules`, which #588 made the one self-containment walk for host packs and `dist` — rather than adding a second validation path for package output; the load scanner itself is one leaf module read by `AB6005` and the prepack gate alike.

## Code

- `packages/agent-bundle/src/build/module-loads.ts` (new leaf; imports only `../core/digest.ts`): `scanModuleLoads(source, { sha256? })` returns every load and loader reference of one source in order — `ModuleLoad = LiteralModuleLoad | ComputedModuleLoad | LoaderReference`, each carrying `form` (`'require' | 'require.resolve' | 'createRequire' | 'createRequire.resolve' | 'bound-loader' | 'bound-loader.resolve' | 'import.meta.resolve'`) and the `loader` as written; plus `quotedLiteral` and `decodeLiteral`. It is the scanner moved out of `pack-dependencies.ts` (`literalLoad`, `computedLoad`, `loaderReference`, `loaderBinding`, `factoryNames`, `loaderNames` are deleted there), rewritten as one token-aware pass: `codeOnly` — the comment/string-blanking pre-pass the old `loaderReference` needed — is deleted, since comments, strings, templates, and regex literals are now stepped over as tokens. Both production callers import it in this change: `pack-dependencies.ts` keeps `moduleLoads(source, sha256?)` as a thin combinator (`complete` = lexed ∧ every dynamic import literal ∧ every load `kind === 'literal'`; `specifiers` = the lexer's literal imports + the literal loads' decoded specifiers) and its `declarationSpecifiers` keeps using `quotedLiteral`/`decodeLiteral` from the leaf; `validate-artifact-modules.ts` reads the loads next to the import records. `DigestCache<T>` (bounded FIFO by insertion, limit in the constructor, `get`/`set`) is extracted to `core/digest.ts`; `module-imports.ts` replaces its inline `importsByDigest`/`importsByDigestLimit`/`remember` with `new DigestCache<readonly ModuleImport[]>(512)`, and the leaf holds `new DigestCache<readonly ModuleLoad[]>(512)` keyed by the bare sha256 — the same "digest of the bytes just read" rule as the import cache.
- `packages/agent-bundle/src/build/validate-artifact-modules.ts`: after the import loop, `scanModuleLoads` on the same bytes and digest; an exhaustive `switch` on `load.kind` with a `never` default — `literal` goes through `resolveJavaScriptImport`, which gains a `via` option (the rendered call) appended to each of its existing messages, and a resolved module is walked by `validateModule`; `computed` and `reference` produce the two new messages. Prebuilt paths still return before any read; the load scan runs for bundles (`lexed`) and parsed generated modules alike, independent of the syntax-check level. Same `graphDiagnostic`, same recovery constant, `reportedRoot` untouched.
- `packages/agent-bundle/src/core/dependency-manifest.ts`: `dependencyManifestPath` is the ancestor `node_modules` walk only; the ``createRequire(join(packageRoot, 'package.json')).resolve(`${name}/package.json`)`` attempt and the `node:module` import are gone. Why: that was the one computed load the audit found in generated code — this module is bundled into every consumer executable that imports `agent-bundle/serve-app-command`, so under the new rule it would fail the artifact build of every project that uses `spawnServeApp`. The walk is Node's ancestor chain done by hand (it was already the fallback whenever a package's `exports` hid `package.json`), covers the same hoisted layouts, and the build-time caller (`declaredDependencyRoots` in `build/rslib.ts`) realpaths the result, so a pnpm symlink is handled where it matters; `locateFrameworkCli` resolves `bin` beside the manifest, which works through the link. Behaviour delta, stated in the changeset: `NODE_PATH` and Node's global folders are no longer consulted. Docblocks in `dependency-manifest.ts` and on `locateFrameworkCli` (`serve-app-command.ts`) say what the code does.
- `packages/agent-bundle/src/build/pack-inventory.ts`: `AB7014` recovery and docblocks reworded — the build inlines every dependency into emitted modules, so load evidence, like import evidence, can only come from a prebuilt payload module or other packed JavaScript outside the artifact and `dist` (and an install script's inline `node -e` program). `package-build.ts`: comment only.

## Fixture proofs

<TBD: confirm the names below against the final tree and paste per-file counts (`n/n`).>

`packages/agent-bundle/tests/module-loads.test.ts` (new, unit) — <TBD: n tests>
- `scanModuleLoads reports a literal load` — every form (`require`, `require.resolve`, direct / qualified / aliased `createRequire` and its `.resolve`, bound loader and its `.resolve`, `import.meta.resolve`) with the decoded specifier (`"\x6ceft-pad"` → `left-pad`)
- `scanModuleLoads reports a computed load` / `… a loader passed on as a value` — the computed and reference shapes listed under Rule
- `scanModuleLoads reports nothing` — comments, strings, templates, regex literals, member calls, `#require`, `__webpack_require__`, `require_fast_uri()`, `typeof require`, `path.resolve`, `Promise.resolve`, `require(id) {`
- `decodeLiteral`, `quotedLiteral names its groups and numbers them 1 and 2 when it opens the expression`, `remembers loads by digest so the same bytes are scanned once per process`
- <TBD: the 56 behaviour fixtures ported from the prototype — 56/56>

`packages/agent-bundle/tests/validate-artifact-modules.test.ts` (new, unit) — <TBD: n tests>
- `names dist paths through reportedRoot the way the package build does`
- `accepts Node built-ins loaded through every resolver, under both spellings`
- `resolves a relative literal load inside the tree and walks the target`
- `reports a relative load whose target is missing, or a JSON target not listed as valid`
- `never scans a prebuilt payload module, even one a compiled module loads`
- `finds the same loads whether a module is lexed as a bundle or parsed in full`
- `reports loads and imports of one module in source order`
- host-pack layout: `leaves a prebuilt payload module opaque to the load scan while a copied module is parsed`
- <TBD: the exact-message cases — `… uses unsupported specifier "left-pad" in require("left-pad").`, `… in load("left-pad"), a createRequire(…) loader.`, `… in import.meta.resolve("right-pad").`, `… loads a non-literal specifier through require(…).`, `… passes require on as a value instead of calling it.` — name the tests>

`packages/agent-bundle/tests/package-build.test.ts` (integration) — <TBD: n/n>
- `fails the package build with AB6005 when node-commonjs externals reach dist through the createRequire shim` — `tools.rspack` `externalsType: 'node-commonjs'`: the build rejects with `AB6005` on `dist/bin/<name>.js` naming the bound-loader call, no `dist` left behind
- `fails the package build with AB6005 when source loads a package through createRequire(), literal or computed`
- the #588 fixtures unchanged: three `AB6005` for externalized imports; `createRequire(import.meta.url)` of a packed file and built-ins under both spellings still pass

`packages/agent-bundle/tests/prepack.test.ts` (integration) — <TBD: n/n>
- <TBD: a `require("x")` inside a comment or string of a packed uncompiled module no longer keeps `x` — `AB7014` reported; name the test>; the recovery assertion (`toContain('devDependencies')`) holds against the reworded string; the #588 `fails prepack with AB6005, never AB7014, …` fixture unchanged

`packages/agent-bundle/tests/pack-dependencies.test.ts` — <TBD: unchanged / n/n>

`packages/agent-bundle/tests/dependency-manifest.test.ts` (new, unit) — 4 tests <TBD: 4/4>
- `finds the manifest under the package root`, `walks up to the ancestor node_modules where hoisting placed a scoped package`, `returns undefined when no ancestor node_modules has the package`, `returns the path through a symlinked package directory (the pnpm layout) and leaves realpath to the caller`

`packages/agent-bundle/tests/serve-app-command.test.ts` — <TBD: `locateFrameworkCli` through the walk, or unchanged>

`packages/agent-bundle/tests/generated-module-loads.test.ts` (new, unit) — the generated-code audit #591 asked for, <TBD: n/n>
- `generated JavaScript loads nothing by a bare package specifier`: every generator that renders a module the plugin build compiles or emits verbatim is rendered with its smallest arguments and scanned with `scanModuleLoads` — `build/entry-shell` (stdio prelude and MCP entry; executable and install bin envelopes; routed CLI bin, render worker, rendered script entry; generated MCP server entry and Flight worker), `build/launch-env-shell`, `adapters/hook-contract` (native and Cursor wrapper codecs; every `planHooks` wrapper through each adapter hook contract, event routes included), `install/surface` (the verbatim installer of every target) — a computed load, a passed-on loader, or a literal bare non-built-in fails the suite and prints the load; `can fail: a createRequire load of a bare package is one offending load` proves the assertion bites. `build/cli-bins.ts` renders no source of its own (it delegates to the `entry-shell` generators above), and the composite root's `bin/`/`mcp/` entries come from the same templates. `pnpm examples:check` building every example (the bundled `serve-app-command` included) is the end-to-end proof: <TBD>.

## Docs

`AGENTS.md` ("Generated plugin output": the walk reads every way a module loads another; the hatch cannot keep a dependency external in any emitted form; `AB7014` evidence comes from files `AB6005` never walked), `docs/diagnostics.md` (`AB60xx` row, `AB7014` row, evidence paragraph), `docs/entry-conventions.md` (`tools` section), `website/docs/{en,zh}/guide/distribution/validation.mdx` (the prepack section and the `AB7014` row) <TBD: confirm `website/docs/{en,zh}/reference/configuration.mdx#tools` and `guide/authoring/package-entries.mdx` needed no change>. Source docblocks in `module-loads.ts`, `pack-dependencies.ts`, `pack-inventory.ts`, `package-build.ts`, `dependency-manifest.ts`, `serve-app-command.ts` match. Stale-wording sweep: `bash /tmp/i591/check-docs-wording.sh` (13 patterns, `--patterns` lists each with why it is stale) → <TBD: `No stale #591 wording in scope.`, exit 0>. Changeset: `.changeset/591-ab6005-module-loads.md`, `minor` for `agent-bundle` (builds and artifacts that passed now fail).

## Gates

- `pnpm build && pnpm typecheck && pnpm lint && pnpm test:unit` — <TBD: ✓ / files, tests, failed>
- `pnpm test:integration:run` — <TBD> (includes `artifact-validator.test.ts`, `package-build.test.ts`, `prepack.test.ts`)
- `pnpm test:packed` — <TBD>
- `pnpm examples:check` — <TBD> (audiobook-curator and host-test are the `dist/bin` consumers; every example that imports `agent-bundle/serve-app-command` bundles the rewritten `dependency-manifest.ts`)
- `pnpm docs:site:build` — <TBD> (dead-link, dead-anchor, language parity)
- dead-module check: `git grep -l 'module-loads' -- ':!repos'` → <TBD: `pack-dependencies.ts`, `validate-artifact-modules.ts`, tests, docs>; `git grep -l 'codeOnly' -- ':!repos'` → <TBD: none>
- `bash /tmp/i591/check-docs-wording.sh` → <TBD: exit 0>

## Self-review

Reviewer: `change-risk-reviewer` on `gpt-5.6-sol-medium` (fallback `generalPurpose` on the same model), prompt at `/tmp/i591/reviewer-prompt.md`, against `origin/main`, asked for concrete merge risks only and specifically: false positives of the scanner on real bundler output (ajv codegen strings, express docblocks, `#require` private methods, `__webpack_require__`, Rspack's `__rspack_createRequire_require` shim loading built-ins), the `reference` rule, the `dependency-manifest.ts` behaviour change for `locateFrameworkCli`, the `AB7014` evidence change, en/zh parity, changeset wording, and stale assertions on the old `AB7014` recovery string.

**Pass 1 — <TBD: n findings>.**
- <TBD: finding> — Disposition: <TBD: fixed in `<sha>` | dismissed: reason>
- Verified with no issue by the reviewer: <TBD>

**Pass 2 (after `<sha>`) — <TBD: disposition accepted; findings>.**


ScriptedAlchemy (commented):
Do not merge this yet. Two open P1 findings are substantive: template-expression loads must be scanned, and replacing Node resolution with an ancestor `node_modules` walk breaks Yarn PnP. The PR body also still contains many `<TBD>` verification placeholders.

Architecturally, keep the goal narrow: prove emitted modules are self-contained. Avoid turning `module-loads.ts` into a general JavaScript semantic analyzer. The more it tries to reason about passed-around `require` values, aliases, lexical forms, templates, regexes, shadowing, etc., the more likely this becomes a fragile parser maintained in parallel with Rspack/Node syntax. Prefer reusing a parser/AST already available in the build toolchain if practical, or constrain the scanner to the exact emitted forms the framework/bundler can produce and validate those with generated-output fixtures.

At minimum add explicit tests for lexical shadowing (`function f(require) { require('x') }`, local `const require = ...`), template substitutions, nested template substitutions, optional chaining/member variants, and minified Rspack output. AB6005 should reject unresolved emitted dependencies, not valid user identifiers that happen to be named `require`.

ScriptedAlchemy:
Closing unmerged — superseded by #619 (owner decision, 2026-09-05 08:33). Nothing from this PR lands; `main` is untouched by it.

> The whole approach is the wrong layer — a meta-framework must not reverse-engineer emitted JavaScript to discover what its own bundler did; Rspack already knows module identity, resolved resource, external modules, dependency type, issuer, chunk membership, runtime requirements. Self-containment is proven from the compiler's module graph and external report, carried in the Artifact IR, and checked by the validator — not inferred by re-parsing generated JavaScript. The scanner also forced a runtime rewrite (`dependency-manifest.ts` lost Node/`createRequire` resolution and broke Yarn PnP): a validator implementation detail must never dictate runtime architecture.

The two P1 threads (template-substitution loads; the PnP-breaking `node_modules` walk) are answered by that decision rather than by further scanner fixes: both symptoms disappear when the evidence comes from the compilation instead of the emitted text. The behavioral tests this PR proved out are listed in #619 under "Behavioral tests to preserve from #602" and salvaged for the implementer.

Issue #619, opened 2026-09-05
Compiler evidence: prove self-containment from the Rspack module graph and Artifact IR; delete the generated-JS load scanners (replaces #591/#602)


ScriptedAlchemy:
Replaces #591 and PR #602 (closed unmerged, owner decision). Shared-compiler-policy step of #592; #604 (Artifact IR / manifest v2) records the compile provenance this issue produces; #555 is the composite root the compiler serves.

## Why (owner rationale, 2026-09-05)

The approach in #591/#602 was the wrong layer. A meta-framework must not reverse-engineer emitted JavaScript to discover what its own bundler did. Rspack already knows module identity, resolved resource, external modules, dependency type, issuer, chunk membership, and runtime requirements. A bespoke scanner over emitted output (`require(…)`, `createRequire(…)`, bound loaders, aliases, templates, regex literals, shadowing, minified forms) is a second JavaScript parser maintained in parallel with Rspack and Node syntax, and it is fragile by construction.

**Dependency-direction lesson from #602.** The scanner flagged Agent Bundle's own `createRequire(…).resolve(…)` in `src/core/dependency-manifest.ts`; the runtime was rewritten to an ancestor `node_modules` walk to satisfy the scanner; Yarn PnP broke (`locateFrameworkCli` → `framework-not-installed`). A validator implementation detail must never force runtime architecture. The direction is: runtime requirements → the compiler resolves them → the validator checks the compiler's evidence.

## Correct flow

```
source
  → Agent Bundle compiler service
  → Rspack compilation / module graph
  → external / module resolution report
  → Artifact IR
  → self-containment validation
  → emit artifact
```

Not: generated JS → reparse → infer → `AB6005`.

## Compiler API

```ts
compiler.compile({ entry, runtime: 'node', selfContained: true })
  → CompileResult { assets: AssetIR[]; modules: ModuleIR[]; externals: ExternalIR[]; diagnostics }
```

`selfContained: true` means any runtime dependency not explicitly permitted by the artifact contract is a compiler error. Enforcement is one loop over compiler evidence:

```ts
for (const external of compilation.externals) {
  if (!isAllowedBuiltin(external.request)) diagnostic('AB6005', …);
}
```

Deterministic, and indifferent to whether Rspack emits `require(...)`, `__rspack_createRequire_require(...)`, or any other shim: the external is known as an external, not inferred from its emitted spelling.

## Closed-world by default

[...]

PR #604, opened 2026-09-05
feat(manifest): authoritative agent-bundle.manifest.json v2 — consumers read the composite root through the manifest (#592 step 3, #555 W2/S3)

ScriptedAlchemy:
Implements **#592 step 3 — the authoritative manifest** together with **#555 W2/S3 — consumers read the composite root through the manifest**. Design comment: https://github.com/ScriptedAlchemy/agent-bundle/issues/592#issuecomment-5549832442. Owner guidance from the 08:30 review is implemented, not noted — see "Owner review (08:30)" below.

## What changes

`agent-bundle.manifest.json` is `manifestVersion: 2`: closed keys, canonical `stableJson` bytes, every path root-relative POSIX, and it is **serialized from the compiler's own IR** (`NormalizedPlugin` → `CompiledRouteGraph` → `TargetArtifactPlan`s) rather than reconstructed. The host documents (`.claude-plugin/plugin.json`, `.codex-plugin/*`, `.cursor-plugin/plugin.json`, `.mcp.json`, `hooks/hooks.json`, marketplace catalogs) are projections the manifest points at; nothing reads them for identity any more.

### Public contract (durable)

| section | fields |
|---|---|
| `manifestVersion` | `2` |
| `application` | `id`, `name`, `version`, `description?` — identity, once |
| `projections[]` | one row per selected host: `host` (the selected name), `builtInHost?` (the shipped adapter that planned it — identity, never the name), `documents{plugin?,marketplace?,mcp?,hooks?}` pointers, `marketplace{name}?` |
| `routes` | the compiled route graph: `digest`, `servers[]{id,name,mode,routes[]}`, `events[]{…,execution{preflight?,providers[],runtime,fallback}}` (#618: the preflight export and required providers artifact-only inspection needs), `scripts[]`, `providers[]`, `layouts[]`, `cli?{mode,routes[],commands[]}` — `commands[]` carry `options[].aliases` and `projection?{module,mapInput,defaults?,relaxed?}` serialized from `CompiledCliProjection` (#616), `contracts?[]{id,input,origin,routes[]}` (reserved shape for #593) |
[...]

PR #634, opened 2026-09-05
build: AB7014 from consumer-runtime evidence (option A); gate the generated-JS walk on the compile evidence record (#619 steps b–d)

ScriptedAlchemy:
Steps (b) and (d) of #619's reordered PR 3 (owner audit 17:11), rebased onto #638 (step a). The coverage matrix (step c) is posted on #619; this PR deletes exactly what it covers and gates the rest.

## AB7014 — option A

`prepack` judges an installed dependency used only when a consumer must have it installed:

- a prebuilt payload declares it in `runtimeDependencies` (`definePrebuilt`);
- a packed `.d.ts` references it — read with TypeScript's `preProcessFile` (imports, `/// <reference types>`, `#subpath` through the `imports` map), not a regex;
- a consumer-side install script names or runs it (command-position `bin`, `node_modules/<name>/` file, bare `--import`/`--require` preload) — the same grammar that escalates `AB7015`.

A package the build inlined is **not** used; the diagnostic names the `dist` bundles that inlined it, read from `PackageBuildResult.evidence` (#638) rather than a scan. `prepack` no longer reads packed JavaScript at all: `require`/`createRequire`/`import.meta.resolve` literals, `bin`-command strings, inline `node -e` programs, files traced from install scripts, and the "computed load withholds the check" rule are gone (matrix rows 8–14). `pack-dependencies.ts` loses ~660 lines.

## Scanner deletion, gated on the record

- **Deleted**: `rslib.ts` `assertNoResidualReservedImports` — a reserved specifier kept external is a `package` `ExternalIR` and fails compile-time `AB6005`; `guardReservedExternals` and `reservedAliasViolation` cover the function-form and alias routes (row 7).
- **Gated**: the emitted-module walk (`validate-artifact-modules.ts`) stays, keyed on `provenModules`. A manifest `bundle` file is proven when the artifact's `agent-bundle.compile-evidence.json` is `AB6039`-clean and `coverage.rewritable` is false; a proven file is lexed, not parsed, and every import it still carries is held to the record: a Node built-in or one of the file's recorded externals passes, a non-literal `import()` fails (the form Rslib leaves verbatim), and any other literal request fails too — the compiler neither bundled nor recorded it, which is what an `rspackIgnore`/`webpackIgnore` import looks like (Rspack leaves it verbatim with no module, external, warning, or dependency object). A matching digest proves the bytes are the compiler's, never that every import was resolved. Every other module — no record, a hatch build, `install.mjs`, copied scripts — keeps the full acorn parse and import resolution (rows 4–6). `ValidateArtifactOptions.bundleSyntaxCheck` / `bundleSyntaxCheckFor` are removed; the record decides.
- **Kept, with a note**: the non-literal `import()` finding for every module, and the walk over JavaScript the framework did not compile. What still needs a replacement is recorded in the matrix and in `docs/entry-conventions.md`.

The package build applies the same gate to `dist`: proven unless a `tools` hatch ran.

## Tests

- `build.test.ts`: a compiled script with `import(name)` fails `build()` with `AB6005 … has a non-literal dynamic import` (compiler evidence cannot see it; the walk does); `import('left-pad' /* rspackIgnore: true */)` and ignored missing/outside/sibling relatives fail `AB6005 … loads "<request>", which the compiler neither bundled nor recorded as an external`; a relocated artifact (source deleted) validates from its record alone and still reports those; hatch build parses in full.
- `package-build.test.ts`: the same ignored-import and expression-import fixtures fail the `dist` walk and leave `dist` unpublished; `evidence.assets[].packages` attributes `node_modules`, linked, and nested-linked dependencies.
- `artifact-validator.test.ts`: walk cases restored; record fixtures for no record / covered / `rewritable: true` / malformed.
- `prepack.test.ts`: inlined-dependency `AB7014` names the bundles; declaration, install-script, and `runtimeDependencies` acceptance; compile-time `AB6005` wins over `AB7014` when only a `dist` bundle imports a declared dependency.
- `generated-module-evidence.test.ts`: the framework's own runtime modules resolve no package (the process-dependency record is empty).

## Docs / changeset

`AGENTS.md` ("Proof is bytes and processes" → compiler evidence + packed-process tests), `docs/diagnostics.md` (`AB6005`, `AB7014`, `AB6039` gate note), `docs/entry-conventions.md`, website en+zh validation, architecture, package-entries, mcp, targets-artifacts. One `minor` changeset (`AB7014` evidence changes; internal `bundleSyntaxCheck` removal is not release-noted).

## Deslop

Fable-5.1-high, rebase pass: one `validateCompileEvidence` (reads, parses, re-judges, and derives `provenPaths`) instead of two readers emitting `AB6039` twice; `publishedPrefix` reused for the `dist` walk's `reportedRoot`; duplicate imports merged; `AB6039` message forms aligned with #638.

## Self-review

Reviewer: `gpt-5.6-sol-medium` (change-risk pass over the diff vs `origin/main`, 4 findings), all fixed in 9e33a0c:

1. **Workspace-linked dependencies lost `ModuleIR.package`** (`rslib.ts` only recognised `node_modules` segments; Rspack records symlinked workspaces at their real path, so `AB7014` could not name the bundle that inlined them). Fixed: `declaredDependencyRoots` now returns real-root → package name and `packageNameOfResource` consults it; `package-build.test.ts` adds a workspace-link fixture asserting `evidence.assets[].packages` attributes `linked-dep`.
2. **No real-build test that the pre-manifest pass engages `provenPaths`** (the distinguishing test used synthetic evidence). Fixed: `build.test.ts` builds a real artifact, re-writes a compiled bundle with a syntax error and re-signs manifest + record, then asserts the record-proven artifact passes the lexed check and the same artifact with `coverage.rewritable: true` fails `AB6005`.
3. **Docs claimed four `AB7014` evidence sources**; `pack-inventory.ts` consumes three and the "process-dependency record" is empty. Fixed across `AGENTS.md`, `docs/diagnostics.md`, website en+zh validation/architecture/targets-artifacts.
4. **Changeset presented the internal `ValidateArtifactOptions.bundleSyntaxCheck` removal as a public breaking change.** Fixed: clause removed.

Second pass (`gpt-5.6-sol-medium`, 2 findings), fixed in the follow-up commit:

5. **Nested linked dependency roots attributed to the parent** (`packageNameOfResource` took the first containing root). Fixed: deepest containing root wins; `package-build.test.ts` adds `linked-nested` inside `linked-dep`'s real root and asserts all three packages (fails without the fix).
6. **Two English passages still said "four" evidence sources** (`docs/diagnostics.md` AB7014 row, `validation.mdx`). Fixed; zh already said three.

Owner re-review (19:56, at `6a1b3b8`): P1 — a proven file skipped literal-import validation, so an `rspackIgnore`/`webpackIgnore` import escaped. Fixed in `2b3f21e` (see the dispositions comment and the [negative-test matrix on #619](https://github.com/ScriptedAlchemy/agent-bundle/issues/619#issuecomment-5554539696)).

Third pass (`gpt-5.6-sol-medium`, 1 finding), fixed in `2014721`:

7. **`artifact-validator.test.ts` still expected a proven bundle's bare specifier and dangling sibling to pass** — the synthetic test encoded the flaw the owner found. Fixed: those are now expected as unaccounted `AB6005`; a recorded `artifact-relative` sibling and a built-in in a proven bundle are shown to pass, and the same sibling without its record entry is shown to fail. Full integration pool: 1127 passed.

Fourth pass (`claude-fable-5-1-thinking-high`, 3 findings), `8cd6d02`:

8. **`docs/diagnostics.md` said the pre-manifest pass validates only filesystem entries and JSON.** Fixed: it also reads the record against the planned file table (`AB6039`) and walks the JavaScript modules.
9. **Changeset and `docs/entry-conventions.md` overstated the function-form reserved-external path** (`guardReservedExternals` intercepts it first; compile-time `AB6005` is the backstop for what reaches the module graph). Fixed in both.
10. **A compiled entry's own `createRequire(…)('pkg')` / `import.meta.resolve` has no declaration path, so `AB7014` fires for it.** Dismissed: this is option A as the owner settled it (bundled ≠ used; the closed-world contract says a generated executable bundles everything but Node built-ins, so a resolution-only load of an installed package is outside the contract), the form is listed in `coverage.unobserved`, matrix row 9 records it, and the owner asked for no unused seam. A `runtimeDependencies` field on entry configs can be added when a fixture needs it.

Final pass (`gpt-5.6-sol-medium`): the three doc edits verified against the code; one low finding — two test comments in `build.test.ts` still described the removed residual-import scan — fixed in `3111475`. No merge risks remaining.


ScriptedAlchemy (commented):
## Re-audit follow-up: keep the AB7014 improvement; block premature AB6005 scanner removal

Reviewed `5870e90554425c14daf2c9d191fbb4bd5027c8c5` through GitHub source/diff only; no builds were executed for this review.

**The option-A dependency policy is correct: bundled does not mean needed by the consumer. The remaining blocker is deleting validation coverage before its replacement exists.**

### P1 — unobserved dynamic loads still escape compilation evidence

[`src/build/compile-result.ts`](https://github.com/ScriptedAlchemy/agent-bundle/blob/5870e90554425c14daf2c9d191fbb4bd5027c8c5/packages/agent-bundle/src/build/compile-result.ts) still explicitly says expression requests can be left verbatim without a module or external record. This PR removes `validate-artifact-modules.ts`, the package `dist` walk, and the residual-import reader without introducing an equivalent compiler rejection/coverage result for that case.

A fixture such as `export const load = () => import(process.argv[2])` must not move from a rejected/unverified runtime edge to a successful self-containment verdict simply because the external list is empty. Generated-module happy-path tests do not establish that negative-case coverage.

**Before deleting the walks:** add the negative fixtures and establish the supported policy with compiler facilities or explicit declarations. Preserve the existing guard where coverage is still missing. This is not a request to revive #602 or grow another JavaScript parser.

### P1 — artifact-only validation has no persisted replacement yet

The PR body defers the manifest half of #619 steps 4–5 until after #604, while this patch already removes the `validate --artifact` JavaScript walk. Main's `compileRslibSurfaces()` passes only `CompileResult.assets` to surface finishers; a new process reading a copied artifact does not possess the compilation's external/module report.

Keep artifact-only validation coverage until the required evidence is recorded and read, or explicitly split and document a narrower validator contract. Merely leaving `AB6005` in the changeset does not make that check run for a separately loaded artifact. Evidence must be associated with the emitted file inventory/digests and its compiler/policy version; retain missing/outside/unlisted relative-target checks.

### Concrete split / merge guidance

1. The consumer-runtime AB7014 evidence work can proceed independently, preserving declaration, declared-payload, and install-script semantics with behavioral tests.
2. Integrate #604 and the necessary persisted-evidence reader/writer contract.
3. Delete each old validator only alongside the replacement for its demonstrated positive and negative cases. Update AGENTS/docs in that same change.

Keep the framework process-dependency record empty while there is no producer; do not add an unused seam or resurrect the deleted checkout launcher.

Required regression matrix: package external via ordinary import and `node-commonjs`; allowed builtin; valid/missing/outside/unlisted relative output; unobserved dynamic import; declared opaque payload; bundled-only package still AB7014; artifact-only validation after relocation without source. Scanner removal remains pending until this matrix has an explicit disposition.

ScriptedAlchemy:
Converting back to draft: the owner audit on #619 (17:11 UTC) reorders PR 3 — the persisted compile evidence record lands first (#638), and scanner deletion is gated on a coverage matrix. This PR will be reworked on top of #638: it keeps the AB7014 option-A re-derivation and deletes only what the matrix covers.

ScriptedAlchemy:
Dispositions for the owner review threads ([17:25 review](https://github.com/ScriptedAlchemy/agent-bundle/pull/634#pullrequestreview-5122215761) at `5870e90`, the 19:55 blocker, and the 19:56 re-review at `6a1b3b8`), against `2b3f21e`:

**P1 — unobserved dynamic loads escape compilation evidence.** Resolved as the re-review records: the walk is not deleted. Every emitted module is still read; a module the record does not prove (no record, `AB6039` findings, `coverage.rewritable`, `install.mjs`, copied scripts) keeps the full acorn parse and import resolution, and a proven module is lexed and fails on `import(<expression>)`. Negative fixture `export const load = (name) => import(name)` is `build.test.ts` *fails the build on an expression import the compiler left verbatim in a compiled script*, plus the new package-output twin.

**P1 — artifact-only validation has no persisted replacement.** Resolved by #638 (`agent-bundle.compile-evidence.json`, `AB6039`), which this PR is rebased onto; `validate --artifact` reads the record from the artifact, not from a compilation. New: `build.test.ts` *validates a relocated artifact from its record alone…* renames the output root away, deletes the project source, and validates — then re-signs a bundle and shows the record-proven bundle is still judged.

**P1 (19:56) — a proven file skips literal-import validation; `import('left-pad' /* rspackIgnore: true */)` escapes.** Confirmed and fixed in `2b3f21e`. Probe first: Rspack 2.2.2 leaves the ignored call verbatim with no module, no external, no warning, and no dependency object on the issuer (only a `category: "unknown"` dependency with no request), so the evidence plugin cannot record it and the first option is unavailable. Second option taken: `accountedRequestsOf(record)` maps each proven file to its recorded external requests, and the walk holds every lexed literal import of a proven file to that set — a Node built-in or a recorded external passes, anything else fails `AB6005 … loads "<request>", which the compiler neither bundled nor recorded as an external; an import the build ignored is a run-time load outside the artifact.` The digest proves the bytes are the compiler's, not that every import was resolved; the walk still resolves nothing the record already judged (no #602 parser). Exact fixture in `build.test.ts` and `package-build.test.ts` (`left-pad`, ignored missing, outside, and existing siblings). Full host/package matrix with test names: [#619 comment](https://github.com/ScriptedAlchemy/agent-bundle/issues/619#issuecomment-5554539696).

**Split guidance (1–3).** Followed in order: #638 is (a); this PR keeps option-A `AB7014` (bundled ≠ used, three evidence sources) and deletes only `assertNoResidualReservedImports` (replaced by compile-time `AB6005` for a `package` external, `guardReservedExternals`, `reservedAliasViolation`) and the `pack-dependencies.ts` packed-JS scanning that option A makes policy; the emitted-module walk is gated, not removed. AGENTS.md and docs updated in the same change.

**Process-dependency record.** Kept empty; no seam added, no launcher revived. Docs now say three `AB7014` sources everywhere (a stale "four" was the previous reviewer's finding).

**Required regression matrix (package external via import and `node-commonjs`; allowed builtin; valid/missing/outside/unlisted relative output; unobserved dynamic import; declared opaque payload; bundled-only package still AB7014; artifact-only validation after relocation).** Every row has a test in the #619 comment; the two `node-commonjs` rows are `compiler-evidence.test.ts` *fails the node-commonjs shim form the same way in a host pack* and `package-build.test.ts` *…when a tools.rspack mutator sets externalsType to node-commonjs*; bundled-only is `prepack.test.ts` *reports a dependency only compiled dist bundles inlined as AB7014, naming the bundles*.

Issue #686, opened 2026-09-06
Stream generated MCP/event Flight bytes before render completion


ScriptedAlchemy:
## Priority and scope

**P2 — production live-render parity.** Follow-up to #600, #643, and #666. Related but separate: #680 owns executable selection; #681 owns retained-history bounds.

Audited commit: `b435f7b9179271cbff81d3e40d14ee342cbd65dd`.

**Evidence level:** the buffering behavior is visible in generated-source code; the proposed end-to-end timing regression has not been run by the reviewer. This is not a claim that Workbench has no streaming infrastructure.

## Problem

The generated route Flight worker collects `renderAgentFlight(...)` through:

```ts
const renderedBytes = new Uint8Array(await new Response(flight).arrayBuffer());
```

It sends the completed bytes after collection. The production Workbench consumer already accepts chunk messages, but cannot display the authored Suspense fallback/replacements incrementally when this producer withholds Flight bytes until completion.

A running status or a separate progress notification does not prove that the authored Flight document is being streamed.

## Source evidence

- [entry-shell.ts](https://github.com/ScriptedAlchemy/agent-bundle/blob/b435f7b9179271cbff81d3e40d14ee342cbd65dd/packages/agent-bundle/src/build/entry-shell.ts): `generatedRouteFlightWorkerSource()` buffers the MCP/shared-event Flight response. Existing rendered CLI/script generation provides a chunked-path comparison.
- [route-invocation-production.ts](https://github.com/ScriptedAlchemy/agent-bundle/blob/b435f7b9179271cbff81d3e40d14ee342cbd65dd/packages/agent-bundle/src/dev/routes/route-invocation-production.ts): `streamFromWorker()` already handles `chunk`, `complete`, and `end` messages through the runtime dispatcher.

## Required correction

Forward Flight chunks through the existing worker/dispatcher transport instead of buffering the entire response first. Preserve request context, provider/state lifetime, byte order, error completion, cancellation, and bounded flow control. Apply the correction to generated event execution where it shares this worker, and verify all consumers of the changed transport.

## Deterministic acceptance

- [ ] Build a real MCP tool with an authored Suspense fallback and a child blocked on an externally controlled test gate.
- [ ] Invoke through Workbench's production MCP surface, not `unit-render`.
- [ ] Assert the actual authored fallback content arrives and is visible **before** releasing the gate; a generic running indicator is insufficient.
- [ ] Release the gate and assert replacement and final document match normal completed execution.
[...]

PR #699, opened 2026-09-06
fix(dev): bound route-invocation render history by count and bytes end to end

### LINKED ISSUE #681: Bound Workbench producer / completed-envelope render history end to end
## Priority and scope

**P2 — memory and response-size reliability.** Follow-up to #666 and #600. Coordinate with production streaming #686, but keep streaming latency and retained-history bounds as separate acceptance criteria.

Audited commit: `b435f7b9179271cbff81d3e40d14ee342cbd65dd`.

Origin: [owner-requested deep-dive, 2026-09-06](https://chatgpt.com/c/6a9b952d-e994-83ea-b7e3-03f9065dde07). Expanded with the complete producer/final/history findings.

**Evidence level:** source inspection establishes that the replay limit does not bound all retained invocation data. The reviewer has not reproduced an out-of-memory failure or claimed that every renderer has no other internal limit.

## Verified gap

The 256-render-event bound applies to one live replay buffer, not the whole invocation lifecycle:

- [route-invocation-production.ts](https://github.com/ScriptedAlchemy/agent-bundle/blob/b435f7b9179271cbff81d3e40d14ee342cbd65dd/packages/agent-bundle/src/dev/routes/route-invocation-production.ts): `renderCompiled()` pushes render events into an array and returns that collection.
- [route-invocation-child.ts](https://github.com/ScriptedAlchemy/agent-bundle/blob/b435f7b9179271cbff81d3e40d14ee342cbd65dd/packages/agent-bundle/src/dev/routes/route-invocation-child.ts): the completed child result carries its event collection through IPC.
- [route-invocation-service.ts](https://github.com/ScriptedAlchemy/agent-bundle/blob/b435f7b9179271cbff81d3e40d14ee342cbd65dd/packages/agent-bundle/src/dev/routes/route-invocation-service.ts): successful invocations retain `events: child.events`; completed history and the final stream envelope retain that result. `#publishStream()` only evicts live render messages.

The final envelope can therefore restore the history that replay eviction discarded. Limiting the number of completed invocations does not establish a bound on retained events or bytes inside each invocation.

## Required behavior

Use one documented retention policy across producer collection, child IPC, live replay, completed invocations, history reads, final stream envelopes, and browser replay. Keep the latest/final document separately from disposable intermediate history. Preserve outcome, correlation, and an explicit truncation indication.

Address retained bytes as well as event counts. Handle oversized final documents separately and explicitly; do not silently truncate valid results or introduce an unjustified universal plugin-result limit merely to cap intermediate history.

## Acceptance

- [ ] Produce substantially more than 256 events through an actual compiled child, not only a mocked `renderChild` callback.
- [ ] Assert producer-retained events and completed `invocation.events` remain bounded.
- [ ] Assert GET/read/history and final-stream responses do not rehydrate the full discarded history.
- [ ] Preserve a single truthful truncation indication and the correct final document/outcome/correlation.
- [ ] Exercise retained-byte behavior with large intermediate snapshots, not only many tiny events.
- [ ] Reconnect after eviction and verify a coherent latest document or an explicit replay limitation.
- [ ] Cancel after the shell event has been evicted and after a long progress sequence; retain intelligible partial evidence and correct cancellation status.
- [ ] Error, cancellation, and completion paths release disposable producer/IPC buffers and preserve process-tree cleanup.
- [ ] Preserve existing live replay behavior; a replay-only test is not sufficient proof of end-to-end bounds.

## Test owners

- `packages/agent-bundle/tests/route-invocation-service.test.ts`
- `packages/agent-bundle/tests/route-invocation-dev-server.test.ts`
- `packages/workbench/tests/invocation-model.test.ts`
- `packages/workbench/tests/audiobook-curator.acceptance.e2e.test.ts`

## Non-goals

No durable trace database, remote telemetry, new orchestration layer, or changes to plugin business behavior. Executable selection is #680; production Flight delivery is #686.

ScriptedAlchemy:
Closes #681

## What

One retention policy for an invocation's render history, applied everywhere it is kept:

- `RENDER_EVENT_RETENTION` (`maxEvents: 256`, `maxBytes: 1 MiB` of JSON) and `retainRenderEvent` / `retainRenderEvents` live in `route-invocation-result.ts`, re-exported from `contracts/invocations.ts` for the Workbench. Oldest-first eviction; the newest event is never evicted, so a completed stream always keeps its `complete` event whatever its size. `document`, `result`, `outcome`, and correlation identifiers sit outside the window and are never truncated (no plugin-result limit is introduced).
- **Producer** (`renderCompiled`) retains in place while streaming and reports `evictedEvents`.
- **IPC**: the child reply carries only the retained window for both production and `unit-render` surfaces.
- **Service**: the live-replay record holds a `RetainedRenderEvents` window plus `latestDocument` tracked outside it; the completed envelope re-applies the bound and carries `evictedEvents` (absent when nothing was evicted). Cancelled and failed runs keep the retained window and the latest document, even after the event that carried it was evicted. `invocationSummary` omits `evictedEvents`.
- **History / reads / final stream message** all serve that bounded envelope, so nothing rehydrates the discarded history.
- **SSE replay** (`#stream`) is paced by socket drain instead of being burst into the live-consumer queue — a reconnect whose retained window exceeded the 256 KiB / 128-record queue limits previously destroyed the response (`TypeError: terminated`), which made the reconnect-after-eviction acceptance impossible. Live frames arriving behind a draining replay are held to the same 256 KiB / 128-record bound on the *current* backlog.
- **Workbench**: the live buffer uses the same primitive (`state.retained`), the client schema accepts `evictedEvents`, and the raw view says how many events were evicted.

Replay order is now: `truncated` marker (if any), the retained render window, kernel trace messages, `final`. No consumer depended on trace/render interleaving.

## Tests

- `tests/route-invocation-retention.test.ts` (unit, no sleeps): count-bound with outcome/correlation kept and history ring eviction; byte-bound with 128 KiB snapshots; the newest event survives at any size; failed run keeps retained history and summaries omit `evictedEvents`; a full 256-render replay served through `RouteInvocationRoutes` over HTTP; reconnect after eviction replays `truncated` + bounded window with no `final`, then cancellation keeps `evictedEvents`, correlation, and the latest document after the shell was evicted.
- `tests/route-invocation-dev-server.test.ts`: new case through a real compiled CLI child (`cli:flood`, 300 Suspense boundaries settling one per macrotask). Completed envelope, `GET` read, and post-completion stream replay are bounded by bytes (well under 256 events) and agree on `evictedEvents`; the final document and `result` are intact. A gated run: live stream past eviction, reconnect replays `truncated` + window only, cancel behind the gate yields `cancelled` with no `outcome`, bounded events with the shell evicted, `evictedEvents`, and the latest document.
- `packages/workbench/tests/invocation-model.test.ts`: count eviction reports `retained.evicted`; byte eviction with 128 KiB snapshots.

## Local gate (branch contains `origin/main` @ 5fc8de37c9; final tree 3bb5240933)

```
pnpm build           → 0
pnpm typecheck       → 0
pnpm lint            → 0
pnpm test:unit       → 0  (4400 passed, 6 skipped)
pnpm docs:site:build → 0  (0 broken links / 28305 anchors)
integration pool (route-invocation-dev-server, trace-dev-server, dev-workbench, workbench-surface-dev-server,
  lifecycle-replay-dev-server, dev-web-host-routes, hook-receipt-pipe, mcp-session-service, host-mcp-proxy,
  mcp-probe-dev-server, host-discovery-dev-server, dev-contract-adoption, target-hook-contract, cli, eval-workbench,
  native-host-sessions, audiobook-curator.acceptance.e2e, examples-real.e2e, overview.e2e)
                     → 0  (131 passed, 0 skipped)
```

One earlier full-suite run hit a load-induced 30 s timeout in an unrelated child-spawning test (`install.test.ts`, then `route-invocation-service.test.ts` "resolves a `.js` import"); both pass standalone and the final full run above is clean.

Deslop: Claude, 3 edits (dropped the redundant `events` alias from the Workbench running state; early return in the child `render`; replaced `findLast` with the retained tail).

## Coordination

#686 owns streaming; retention is settled here — see the comments on #681 and #686. Any chunked Flight delivery should feed `publishRender` one event at a time and reuse this primitive; the SSE route keeps its live-consumer limits.

## Self-review

Reviewer: `gpt-5.6-sol-max` (`change-risk-reviewer`), three passes.

1. Count-bounded SSE replays destroyed by the live counter — **fixed** (only live arrivals count; HTTP replay test added).
2. `invocationSummary` leaked `evictedEvents` into strict Workbench list decoding — **fixed** + test.
3. `JSON.stringify(...).length` is UTF-16 code units, not UTF-8 bytes — **dismissed**: the bound protects the dev server's heap, and V8 stores non-Latin1 strings two-byte, so code units track resident size more closely than UTF-8; the unit is stated on the constant.
4. Failed envelopes discarded retained evidence while the stream replayed it — **fixed** (`retainedHistory(record)` shared by cancelled and failed paths) + test, docs, changeset.
5. `unit-render` still collects the full stream in the harness before the IPC bound — **dismissed**: `renderRouteEvents` is the public test-harness contract and returns every event by design; IPC is bounded here and the runtime caps a stream at 10 000 events.
6. (pass 2) Live backlog behind a replay used a cumulative counter with no byte bound — **fixed** (pre-encoded frames; current-backlog records and bytes, decremented as pumped).

Pass 3: no remaining merge risks.


ScriptedAlchemy:
Closing as a duplicate of #715, which also fixes #681.

Both PRs bound producer, IPC, replay, completed/cancelled/failed envelopes and history by event count and bytes and pin the final outcome/correlation. #715 survives because it goes further on the producer side: the child no longer carries any `events` array over IPC (each event is published as it happens and the send is awaited, so the channel paces the producer), the newest document-bearing event is pinned so a replay after eviction still folds to a coherent document, the truncation account (`retention`) is the one truthful record, and it forwards compiled routes' progress reports as render events, which is what lets a compiled MCP tool exceed the window at all.

Ported from this PR to #715 (a643d08dbf): the real-compiled-child `cli:flood` dev-server test (byte-bound snapshots, GET/replay parity, mid-run reconnect after the shell was evicted, cancel after eviction), the pinned-newest module test, the history-ring eviction test, and the full-window HTTP replay through `RouteInvocationRoutes`. Retention module: `packages/agent-bundle/src/dev/routes/route-invocation-render-history.ts` (256 events / 2 MiB).

PR #754, opened 2026-09-07
docs: complete convention-first guides, targets, and capability navigation (EN/ZH)

ScriptedAlchemy:
Completes the convention-first onboarding path on the docsite in both locales: one authoritative
Quick start that adds a tool by writing a route file, a capability map that separates compiled
output from client behavior, and reference/troubleshooting pages that send an author to the
existing generated matrices instead of a second hand-maintained one.

Pages added: `guide/start/capabilities`, `guide/authoring/reuse-framework`,
`guide/development/troubleshooting` (en + zh, with matching `_meta.json` navigation). Pages
rewritten: `guide/start/{index,quick-start,project-structure}`, `guide/distribution/index`,
`reference/{index,targets-artifacts}`.

Docs-only: no publishable package changes, so no changeset (`skip-changeset`).

## Local gate (authoritative)

Run on `867d3787fb`, which contains current `origin/main`:

| Command | Result |
| --- | --- |
| `pnpm build` | pass |
| `pnpm typecheck` | pass |
| `pnpm lint` | pass |
| `pnpm test:unit` | pass, 0 failures |
| `pnpm docs:site:build` | pass — locale drift, diagnostics coverage, Rspress build, language parity, 0 broken links / 30105 anchors over 35421 internal links in 122 files |

## Tutorial and Twoslash acceptance

Twoslash is a real gate, verified by falsification rather than assumed: with an unresolvable
import in the Quick start's `defineConfig` sample the docsite build fails
(`[@rspress/plugin-twoslash] Twoslash error in code … 2307`, exit 1). All ten twoslash blocks in
this PR compile against the built declarations. (A first attempt at falsification — an extra
key in the `defineConfig` object — correctly did *not* fail, because `defineConfig` is generic
and infers the literal type.)

The Quick start's "add your first tool" journey was executed end to end against a copy of the
`mcp-server` template, using the page's exact code:

- `src/mcp/status/tools/hello.tsx` (route file only, no registration edit) → `validate` succeeds.
- `renderRoute('tool:status/hello', { input: { name: 'Ada' } })` → `result` is
  `{ message: 'Hello, Ada.' }`, passing in the template's own `route-unit` pool.
- `hello.cli.ts` with `command: ['hello']` → `build` emits the routed bin, and
  `node artifact/bin/<plugin>.mjs hello --name Ada` prints `Hello, Ada.`
- `agent-bundle --version` and the documented `inspect --artifact` / `--root` exclusivity
[...]

PR #757, opened 2026-09-07
feat(routes): type callers by schema input; per-program AB4834; validate before typecheck (#748, #752)

### LINKED ISSUE #748: [P2] Make generated-route typing reliable for clean checks and every consuming TS project, rather than an existence check in one program
## User problem

The basic promise is that an authored route gives the browser, providers and tests current types. Today that promise depends on a prior build/dev/validate command and on correctly including `.agent-bundle/routes.d.ts` in the actual consuming TypeScript program. A successful unrelated project can mask a missing augmentation.

Audited `14c9822bc6c01d8ff2788454d0e1dc7861bc3816`; source review, no fresh TypeScript invocation. This is a focused authoring follow-up, not a claim that generated typing is unimplemented.

## Concrete evidence

- [routes/typegen-program.ts](https://github.com/ScriptedAlchemy/agent-bundle/blob/14c9822bc6c01d8ff2788454d0e1dc7861bc3816/packages/agent-bundle/src/routes/typegen-program.ts) returns no diagnostic if the file is absent, or if **any one** directly referenced program includes it. For a solution containing `tsconfig.node.json` and `tsconfig.app.json`, the node project can include the declaration while the browser project does not; AB4834 then reports no problem even though App route IDs/results fall back to weaker types in the browser program. Nested references are not traversed.
- [MCP starter package scripts](https://github.com/ScriptedAlchemy/agent-bundle/blob/14c9822bc6c01d8ff2788454d0e1dc7861bc3816/packages/create-agent-bundle/templates/mcp-server/package_json) expose plain `tsc` as `typecheck`; the larger `check` script happens to run build first. The generated file is intentionally gitignored.
- [typegen.ts](https://github.com/ScriptedAlchemy/agent-bundle/blob/14c9822bc6c01d8ff2788454d0e1dc7861bc3816/packages/agent-bundle/src/routes/typegen.ts) already owns atomic publication. Reuse that implementation and the canonical graph; no second route parser is required.

## Required simplification

Provide one documented reliable entry for generating current contracts before an isolated typecheck, using existing validation/type generation where sufficient. A lightweight explicit command or shared template script is acceptable; a second compiler is not. Do not force a full production bundle just to learn route IDs. Update the starter so the documented typecheck is correct from a clean checkout and after route changes, not merely after some other workflow warmed generated output.

Make the inclusion diagnostic correspond to the program that actually consumes the framework's augmented contracts. Do not require generated types in unrelated build-only projects. If automatic consumption detection would require invasive source analysis, allow the relevant tsconfig selections explicitly and validate those; do not claim global type safety from an existential check over references.

## Acceptance

- [ ] Clean project: the documented typecheck creates current declarations or stops with an actionable prerequisite error; it must not silently succeed with weakened route types.
- [ ] Add/rename/delete a route and change its schema/provider keys; the next check sees the new contract without restarting a dev server.
- [ ] Solution-style root with separate browser/server configs: inclusion in the server project cannot hide exclusion from the browser consumer.
- [ ] Nested references, extends, excluded declarations and case-sensitive paths are tested; unrelated projects are not forced to import framework types.
- [ ] Invalid route ID/input/result access is tested as a negative compile case in the actual App program.
- [ ] Failed generation never replaces a valid declaration with partial content; stale output is not described as fresh.
- [ ] Static-only plugins remain free of unnecessary route declarations/runtime dependencies.

Document `dev`, `build`, `typecheck` and CI behavior once. Delete consumer-specific shell existence probes when the supported path is adopted. Do not silently edit all user tsconfigs, run user tool/provider code for metadata, or impose a TypeScript major upgrade to solve a generation-order problem.

### LINKED ISSUE #752: [P2] Generate distinct caller-input and parsed-handler types so defaults and transforms work through Apps and test helpers
## Audited finding

Commit: `14c9822bc6c01d8ff2788454d0e1dc7861bc3816`. GitHub source review; the proposed TypeScript/runtime regression fixtures were not executed by this reviewer. This is distinct from #748 (freshness/inclusion of the generated program), #746 (CLI grammar coverage), and completed #593 (local schema imports).

The framework should let an author define one schema and then call the operation naturally. Today its generated public invocation types use the schema's **output** as the caller's **input**.

## Concrete source chain

- [routes/typegen.ts](https://github.com/ScriptedAlchemy/agent-bundle/blob/14c9822bc6c01d8ff2788454d0e1dc7861bc3816/packages/agent-bundle/src/routes/typegen.ts): `RouteContract<InputSchema, ResultSchema>` uses `input: SchemaOutput<InputSchema>`, where `SchemaOutput` reads `_output`. `AgentBundleAppRouteContracts` and the harness register both inherit that same input.
- [app/index.ts](https://github.com/ScriptedAlchemy/agent-bundle/blob/14c9822bc6c01d8ff2788454d0e1dc7861bc3816/packages/agent-bundle/src/app/index.ts): `AppClient.call()` accepts `AppRouteInput<Id>` and `onToolInput` exposes the same contract, although these are messages before server parsing, not already-normalized handler props.
- [routes/public.ts](https://github.com/ScriptedAlchemy/agent-bundle/blob/14c9822bc6c01d8ff2788454d0e1dc7861bc3816/packages/agent-bundle/src/routes/public.ts): `RouteSchemaOutput` is appropriate for handler props, which receive the validated value. It must not be changed blindly for every use.
- [build/entry-shell.ts](https://github.com/ScriptedAlchemy/agent-bundle/blob/14c9822bc6c01d8ff2788454d0e1dc7861bc3816/packages/agent-bundle/src/build/entry-shell.ts): generated MCP invocation receives input the SDK validated; the Workbench validation path applies the route input schema before constructing props. The type contract must correspond to each side of that parse.
- Zod's [input/output documentation](https://zod.dev/basics#inferring-types) and [defaults](https://zod.dev/api#defaults) explicitly distinguish these types.

## Small regression case

```ts
export const inputSchema = z.object({ limit: z.number().default(10) });
```

The valid caller input `{}` should typecheck in `client.call(id, {})`, while the handler should continue seeing a required `limit: number`. Generating both sides from `_output` incorrectly makes the caller fill a field that the schema deliberately defaults.

Where a runtime-supported schema transforms a wire string into a number, the reverse mismatch is more serious: a parsed numeric value can typecheck as caller input although the server expects the original string. Include that case only through the framework's genuinely supported schema path; this ticket does not demand that every arbitrary transform be statically representable in JSON Schema.

## Correction

Carry separate schema-input and schema-output types through the existing generated contract. Reuse the same type-only route imports; no second schema registry, evaluated-discovery phase, or browser copy of Node/Zod modules. Define which public helper accepts raw invocation input and which, if any, deliberately accepts already-parsed props. Maintain that distinction rather than globally renaming `RouteInput` and breaking route execution.

Audit `createAppClient.call`, opening-input listeners, route test helpers, in-memory MCP calls, CLI mapInput and handler props against their actual parse boundary. Results remain the validated result-schema output subject to the existing strict-JSON wire rules.

## Acceptance

- [ ] Defaulted/optional fields may be omitted by an App caller; the handler receives the default and retains useful nonoptional typing.
- [ ] Required fields, unknown route IDs and wrong primitive values remain negative compile cases.
- [ ] A supported transformation/coercion fixture verifies raw input, parsed handler props and result types separately; processing does not apply a transform twice.
- [ ] The actual generated `.agent-bundle/routes.d.ts` is used by the compiler test, not a handwritten stand-in for AppRegister.
- [ ] One compiled MCP/App round trip demonstrates the same behavior as the route helper, with invalid input rejected before side effects.
- [ ] Opening `onToolInput` does not promise server defaults are already present unless the actual host payload contains them.
- [ ] Type-only imports keep server implementations, Node APIs and Zod out of the emitted browser runtime.
- [ ] Structural/non-Zod schema compatibility gets an explicit input/output rule; no silent any/unknown widening merely to silence errors.
- [ ] Update the public type documentation, examples and changeset in the same change.

## User-experience goal

No `as any`, manual default duplication, or separate browser schema is required to call a normal defaulted route. The schema is authored once, and the framework supplies the right view of it at each boundary.

ScriptedAlchemy:
Closes #748. Closes #752.

## What

**#752 — caller input vs parsed handler output.** The generated `.agent-bundle/routes.d.ts` now derives each route's `input` from the schema's `_input` (what a caller sends) instead of `_output`. `createAppClient().call`, `onToolInput`, `renderRoute`, `invokeMcpTool`, and contract-matrix fixtures accept `{}` for `z.object({ limit: z.number().default(10) })` and a string for a `.transform()`ed field; `ToolRouteProps<typeof inputSchema>['input']` keeps the parsed output. A structural schema declaring only `_output` uses it for both (explicit rule, spelled in the generated file). `renderRoute` now parses input through the route's own `inputSchema` at the same boundary the generated Flight worker does — defaults filled, transforms applied once — and rejects invalid input with `invalid-input` before any provider or component runs. `LoadedRouteModule.inputSchema.parse` returns `unknown` (the parsed props are not part of the registration).

**#748 — reliable typegen / per-program AB4834.** `routeTypesProgramDiagnostics` walks `references` transitively and reports `AB4834` once per program that *consumes* the registration (a source file importing `agent-bundle/app`, `agent-bundle/test`, `agent-bundle/eval`, or `@agent-bundle/runtime`) and omits the declaration; build-only projects are left alone; recovery names the include path relative to the offending tsconfig. The documented entry for current declarations before an isolated typecheck is `agent-bundle validate` (~3s, no bundle): the `mcp-server` and `cli-tool` starters' `typecheck` is now `npm run validate && tsc`, and `check` no longer runs validate twice.

## Tests
- `tests/route-caller-input-types.test.ts` (integration): clean generated project with a solution-style root (`tsconfig.app.json` browser program with `lib: DOM`, `tsconfig.node.json` server program). `validate` publishes the declaration; both real tsconfig programs type-check via `ts.parseJsonConfigFileContent`; `client.call('tool:curator/page', {})`, `onToolInput` with optional default, transform input as string; negatives (wrong id, missing required, wrong primitive, parsed-number-as-caller) each one error; dropping the declaration from the browser program alone yields one `AB4834` on that program; built MCP server round trip applies default/transform once and rejects invalid input before the handler runs (invocation log).
- `tests/route-types-program.test.ts`: nested references, `exclude`d declaration, build-only project untouched, recovery path relative to nested tsconfig.
- `tests/route-graph.test.ts`: `_input`-bearing and `_output`-only structural stand-ins.
- `tests/route-unit/render-route.test.ts`: `renderRoute` applies the default and rejects `{ label: 1 }` with `invalid-input`.
- `route-register-typegen`, `generated-route-server`, `route-contract-imports` updated/green; packed scaffold matrix (`mcp-server`, `cli-tool`, `minimal`) green with the new `typecheck` script.

## Docs
en+zh: `guide/development/testing.mdx`, `guide/development/index.mdx`, `guide/authoring/mcp.mdx`; `docs/diagnostics.md` (AB4834); `docs/entry-conventions.md`; starter READMEs. No overlap with #754's paths.

## Gate (current origin/main)
`pnpm build && pnpm typecheck && pnpm lint` ✅ · `pnpm test:unit` ✅ · route-unit pool (90) ✅ · projection pool (191) ✅ · `test:packed:release` scaffold matrix ✅ · full integration pool + `pnpm docs:site:build`: running, results appended below.

Deslop: pending. Self-review: pending.


ScriptedAlchemy:
Codex first-pass findings, addressed locally:

1. **Changeset `(#PR)` placeholder** — fixed in `dbb8408`; the summary now ends with `(#757)`.
2. **AB4834 scanned only the configured root files** — correct: `parseJsonConfigFileContent().fileNames` is the root list, so `files: ["src/index.ts"]` importing a consumer was missed. `typegen-program.ts` now builds the program from those roots (`ts.createProgram`, host declining `node_modules`, `noLib`, `types: []`) and scans `getSourceFiles()`, i.e. the same import closure `tsc -p` compiles; a declaration reached through `/// <reference>` also counts as compiled. Regression test: `route-types-program.test.ts` “follows imports from a narrow root”. `validate` on `examples/host-test` costs ~0.6 s more.

Also fixed from the failed Examples check: `examples/host-test` rendered `slow` with `tickMs: 40` against a schema whose minimum is 50 — a path the runtime rejects; now that `renderRoute` parses input, the test uses `tickMs: 50` (`17924a5`). `pnpm examples:check` passes locally.

ScriptedAlchemy:
Local gate at `f046cf8` (rebased on current `origin/main`): `pnpm build`, `pnpm typecheck`, `pnpm lint`, `pnpm test:unit`, `pnpm test:integration` (whole pool, 1171 passed / 0 failed), `pnpm examples:check` (Examples root cause: `examples/host-test` `slow` probe sent `tickMs: 40` below the tool's `min(50)`; the harness now validates, the test was corrected rather than the check loosened), `create-agent-bundle` packed scaffold e2e, `pnpm docs:site:build` — all green. Independent local review (Fable): nothing blocks.

PR #824, opened 2026-09-17
fix(install): Codex add-only replace; refuse disabled/unknown enablement

ScriptedAlchemy:
Codex replacement currently runs `plugin remove` before `plugin add`. Remove deletes the plugin's settings subtree, so an enabled plugin with a disabled nested MCP server loses that override during an upgrade.

Replacement now uses native `plugin add` without removal, preserving nested settings and refreshing the cached artifact. Disabled or unknown-enablement replacements refuse before mutation (`AB7004`), since native add enables the plugin and the plugin CLI offers no qualified way to preserve its disabled state. Failed add/receipt recovery also avoids removing an existing Codex plugin. Claude, fresh installs and explicit uninstall retain their existing behavior.

The installer does not write or restore `config.toml`. Native inventory is a snapshot: a concurrent edit to this same plugin's enabled flag between list and add is not atomically protected. Unrelated configuration is never rewound.

Validation: independent exact-head review **PASS+NOTES** at `2ebfcd87ba803a4a6b2f80288ebee809d4ae41e8`; isolated actual Codex 0.154 add-only refresh/removal-of-obsolete-files/nested-override proof; enabled/disabled/unknown and add/receipt-failure regression tests; native CI, host-install and Ubuntu/macOS/Windows filesystem checks, build/types/release gates and EN/ZH documentation checks. A patch changeset is included.


ScriptedAlchemy:
## HOLD merge (owner note)

Merge is **held** at `269f875` pending independent review + revised implementation/proof.

Parent constraints (must land before merge):

1. **No whole `config.toml` capture/restore** — that can overwrite unrelated concurrent user/native edits and violates the preservation boundary.
2. **Fail closed before mutation** when existing disabled state cannot be safely preserved (missing/unreadable prior config must not re-enable a disabled plugin).
3. **Native add-only** remains the proved/correct base for nested overrides / version / drift.
4. Restore **only** the exact prior plugin setting that native add changes; prefer a **supported native settings API** with expected-version compare if available.
5. **Do not** overwrite whole user config, add a TOML parser/framework, or invent permissions.

Cloud worker owns the surgical revise + PR body rewrite around the corrected final approach. Independent review + proof required before any merge.

ScriptedAlchemy:
## Independent review: FAIL @ `269f875` — merge still held

Blocking:
1. Whole-config restoration can overwrite unrelated concurrent edits.
2. Treating all read errors as absent can re-enable a disabled plugin.
3. Direct `writeFile` truncates live config and may leave it partial on failure.

Required direction (owner → cloud worker):
- Keep **native add-only**.
- Discovered regression = **enabled replacements with existing nested overrides** → must work natively without remove.
- Restore/correct **only** the prior plugin `enabled` flag via a **qualified scoped native write with conflict detection**; preserve unrelated config + native nested tables.
- If pinned host lacks a qualified settings-preserving update API: **fail closed** before mutation for disabled/unknown-enablement replacement; document that supported boundary; leave those installs unchanged.
- No general TOML parser / new client framework to override user intent.
- Fresh install, Claude, explicit uninstall unchanged.
- Tests: concurrent unrelated edits; config read/write failure; failed-add/receipt recovery.

Merge only after a **revised exact head** passes independent review. PR body must match final behavior.

